PR #1, opened 2020-01-19
Remove the use of Shared from the public API to avoid consumers being required to use unsafe to use the map

jhinch:
My expectation is that a regular user of the map should not be required to use `unsafe`. `Shared` requires that you use unsafe to dereference to the value. Instead, tie the reference to the lifetime of the guard passed into the get method

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,8 +177,6 @@ where
 
     pub fn get_and<R, F: FnOnce(&V) -> R>(&self, key: &K, then: F) -> Option<R> {
         let guard = &crossbeam::epoch::pin();
-        // safety: we are still holding the guard, and saw `v`, so it won't be dropped until the
-        // next epoch after we drop our guard at the earliest.
         self.get(key, guard).map(then)
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,7 +124,7 @@ where
     }
 
     // TODO: implement a guard API of our own
-    pub fn get<'g>(&'g self, key: &K, guard: &'g Guard) -> Option<Shared<'g, V>> {
+    pub fn get<'g>(&'g self, key: &K, guard: &'g Guard) -> Option<&'g V> {
         let h = self.hash(key);
         let table = self.table.load(Ordering::SeqCst, guard);
         if table.is_null() {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -159,14 +159,17 @@ where
...
         let v = node.value.load(Ordering::SeqCst, guard);
         assert!(!v.is_null());
-        Some(v)
+        // safety: the lifetime of the reference is bound to the guard
+        // supplied which means that the memory will not be modified
+        // until at least after the guard goes out of scope
+        unsafe { v.as_ref() }
     }
 
     pub fn get_and<R, F: FnOnce(&V) -> R>(&self, key: &K, then: F) -> Option<R> {
```
jhinch: Arguably, `get_and` doesn't need to exist now

src/lib.rs
```diff
@@ -159,14 +159,17 @@ where
...
         let guard = &crossbeam::epoch::pin();
         // safety: we are still holding the guard, and saw `v`, so it won't be dropped until the
         // next epoch after we drop our guard at the earliest.
-        self.get(key, guard).map(|v| then(unsafe { v.deref() }))
+        self.get(key, guard).map(then)
```
jonhoo: The safety comment above this line can go away now!

jonhoo:
I like this! I'd still like to _also_ provide a version where the caller doesn't need to supply a guard, but this is a good step forward. Could you rebase onto master so that we also get the CI tests I just pushed?

src/lib.rs
jonhoo: Feel free to remove it!

jhinch: I think given your comment with regards to an API which doesn't require a guard to be supplied, I think for now I'll leave it. I have some ideas of how to provide a more flexible API in this regard so I might raise that as a separate PR

jhinch:
Feedback addressed. Changes rebased onto master and builds are passing. Should be good to merge

jonhoo:
Beautiful, thank you!

jonhoo:
Note that I just [changed the license](https://github.com/jonhoo/flurry/commit/3242c25ed8ea75a4eb89efae9e7623f8b7245212) of the repository. Are you okay with this change being licensed under MIT + Apache 2.0?

PR #2, opened 2020-01-19
Add iter, keys, and values with Java tests

thallada:
I was following part of you stream earlier today and decided to try my hand at porting some of the other Java tests from [MapCheck.java](https://hg.openjdk.java.net/jdk/jdk/file/345266000aba/test/jdk/java/util/concurrent/ConcurrentHashMap/MapCheck.java).

The contains_key tests were pretty straightforward since you already implemented `contains_key`. Those are `t4` and `t7`.

During the stream, you mentioned that `NodeIter` is a private Iterator which could be the basis for public iterators on keys, values, and entries. I've done that in a new file `src/iter/iter.rs` with three new structs `Iter`, `Keys`, and `Values` (similar to `std::collections::HashMap`). I added some of my own tests for the new `iter()`, `keys()`, and `values()` methods and ported `ittest1`, `ittest2`, and `ittest3` from MapCheck.java.

To get the values out of the `NodeIter`, I had to add two more unsafe blocks.

I'm still a noob at concurrency and Rust so let me know if there's anything to improve here. I'm using this as a learning experience. Thanks for streaming!

### DIFF
--- a/src/iter/iter.rs
+++ b/src/iter/iter.rs
@@ -1,0 +1,122 @@
+use crossbeam::epoch::Guard;
+use std::sync::atomic::Ordering;
+
+use crate::NodeIter;
+
+/// An iterator over the entries of a `FlurryHashMap`.
+///
+/// This `struct` is created by the [`iter`] method on [`FlurryHashMap`].
+/// See its documentation for more.
+///
+/// [`iter`]: /flurry/struct.FlurryHashMap.html#method.iter
+/// [`FlurryHashMap`]: /flurry/struct.FlurryHashMap.html
+#[derive(Debug)]
+pub struct Iter<'g, K, V> {
+    pub(crate) node_iter: NodeIter<'g, K, V>,
+    pub(crate) guard: &'g Guard,
+}
+
+impl<'g, K, V> Iterator for Iter<'g, K, V> {
+    type Item = (&'g K, &'g V);
+    fn next(&mut self) -> Option<Self::Item> {
+        let node = self.node_iter.next()?;
+        let value = node.value.load(Ordering::SeqCst, self.guard);
+        // safety: flurry does not drop or move until after guard drop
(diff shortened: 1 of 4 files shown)


src/iter/iter.rs
```diff
@@ -0,0 +1,104 @@
...
+
+pub struct Iter<'g, K, V> {
+    pub(crate) node_iter: NodeIter<'g, K, V>,
+    pub(crate) guard: &'g Guard,
+}
+
+impl<'g, K, V> Iterator for Iter<'g, K, V> {
+    type Item = (&'g K, &'g V);
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(node) = self.node_iter.next() {
```
jhinch: You can flatten out this code by using `?`

```
let node = self.node_iter.next()?;
let value = node.value.load(Ordering::SeqCst, self.guard);
// safety: flurry does not drop or move until after guard drop
let value = unsafe { value.deref() };
Some((&node.key, &value))
```

tests/jdk/map_check.rs
```diff
@@ -55,4 +121,15 @@ fn everything() {
     t1(&map, &keys[..], SIZE);
     // get (absent)
     t1(&map, &absent_keys[..], 0);
+    // contains_key (present & absent)
```
jonhoo: Can you place these in the same order as they are in the original Java tests? Just to make referencing those files easier in the future?

jonhoo:
I love this change, thank you! Could you rebase onto master so that you get the CI testing changes I just pushed?

tests/jdk/map_check.rs
thallada: Yep, done.

thallada:
Rebased and added docs. Let me know if this needs anything else!

jonhoo:
Excellent, thank you!

PR #4, opened 2020-01-20
Added a stripped-down implementation of remove plus tests

domenicquirl:
Hi, I'm relatively new to rust but thought this was a nice opportunity to try my hand at some more real-world problems. 

There are some differences to the Java code:
- the Java code falls back on an method `replaceNode` which can also replace the previously stored value and perform a comparison check on the stored value to determine if it should be replaced. The replacement is probably possible to do, but it left it out since it is not part of the `remove` functionality. The value comparison I think would require the map values to be `EQ`, which they are not currently and I didn't want to change the overall contract of the type.
- because of this, the tree bins which do not exist and differences in implementation several cases/checks do not exist in this implementation.

I was also unsure about whether to return the removed value (which I decided to do) and if so, whether it needed to be marked for `defer_destroy`. Lifetimes are still somewhat confusing to me in this scenario. So maybe someone could review that and get back to me?


jonhoo:
I'm a little squeezed for time these days, and this requires a thorough review, so it may take me a little time to get back to you on it. Thank you for contributing though — this is a good change. If someone else finds the time to help review, I'd be very grateful for that!

For some high-level comments:
 - It would be great if you could also port some of the Java tests related to removal.
 - It looks like CI does not pass — we should fix that.
 - You _will_ need to `defer_destroy`. Take a look at what the code in `put` does when it needs to replace a value. For `remove`, you will need to do something similar, though you will _also_ want to `defer_destroy` the node.

Separately, I like the idea of returning `&'g V` from `remove`. We should do the same for `insert`. It would mean that `insert` also has to take a `Guard` rather than make one itself, but I think that's the right thing to do for the time being. We will eventually provide a wrapper that hides the `Guard` too I think. That can all go in a separate PR though!

domenicquirl:
Don't worry, as I said I'm doing this mostly as an exercise for myself. Also, there are apparently still problems with the implementation. I added the Java tests for removal from `MapCheck`, although they are not very parallel, so ironically it is my previous concurrent test that catches most of the errors. I'll need help with these however, though of course not necessarily from you if you don't have the time.

Issues:

- If an element with the given key is found, the `remove` implementation performs `add_count` with a value of `-1` (and a resize hint of `None`, so no resizing). `add_count` then does a `fetch_sub` on the atomic count value and manually recomputes the `-1` on the returned value. This seems to sometimes cause an over- (or under-)flow. However, under all of printing values for debug, loading the count value and casting the returned value back and forth, I cannot reproduce the issue.

- The implementation seems to possibly produce heap corruption across tests from `basic.rs`, which may or may not be related to the cause of the above.

I did add `defer_destroy` for the removed node though, and also included the API change moving from `Shared<'g, V>` to `&'g V`.

jonhoo:
For those with some time to help, or for me later, it would be helpful if you could give the backtrace of _where_ the under/overflow happened in the code. Is it on the `fetch_sub` itself? Could it be that you call `add_count(-1)` even if no element was removed (since it was not found)?

domenicquirl:
As of right now the overflow issue does not reproduce on will. If I hit it again, I will provide the backtrace.

Regarding the other issue:
Based on the output of CI (particularly the failing Linux tests), the double free/corruption seems to stem from somewhere related to `_ZN15crossbeam_epoch8internal6Global7collect...` / `_ZN72_$LT$crossbeam_epoch..internal..Bag$u20$as$u20$core..ops..drop..Drop$GT$4drop`, which to me suggests the issue is related to epoch garbage collection. I find that in testing, not using `defer_destroy` on the removed value [(which happens here)](https://github.com/domenicquirl/flurry/blob/master/src/lib.rs#L936) consistently gets rid of the error. Of course, then the old value might not be marked as garbage.

jonhoo:
Well, if you don't use `defer_destroy`, then the value does not get freed at all, so that would explain that :p It suggests to me that the old value is still sticking around somewhere in the map after you remove it, and then some other thread tries to destroy another instance of the same value later...

domenicquirl:
Found it! I missed changing one static reference to a local one when deciding which of the inline-assigned Java locals were needed. Since the change matched your helpful pointers - thanks a lot! -, I consider the issue resolved. CI is also passing now :)

jonhoo:
Awesome! I'll see if I can find time to review this in not _too_ long :sweat_smile: 

In the meantime, it looks like there is something odd about your commits. Could you try rebasing your changes onto `master`? Also, as mentioned, it'd be great to have some more of the Java tests for `remove` ported! I recently imported some more Java tests into `jsr166/test/ConcurrentHashMapTest.java`, which should be a good starting point (probably start a new file under `tests/jsr166.rs`).

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+            //     `next_table` raw pointer from inside the Moved. how do we know that that is safe?
+            //
+            //     we must demonstrate that if a Moved(t) is _read_, then t must still be valid.
+            //     FIXME cf. put
+            let t = unsafe { table.deref() };
+            let n = t.bins.len() as u64;
+            if n == 0 {
+                break;
+            }
+            let i = ((n - 1) & h) as usize;
```
jonhoo: This should probably use `Table::bin`. I think in general it'd be good to model this function as closely as possible to what is in [the ported code for `put`](https://github.com/jonhoo/flurry/blob/437663f284aa68d01cd806cbbd0a1b2bc7ce8ea9/src/lib.rs#L476) so that we keep the things looking similar.

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+            match *unsafe { bin.deref() } {
+                BinEntry::Moved(next_table) => {
+                    table = self.help_transfer(table, next_table, guard);
+                }
+                BinEntry::Node(ref head) => {
+                    let head_lock = head.lock.lock();
+                    let mut old_val: Option<Shared<'g, V>> = None;
+                    let mut validated = false;
+
+                    if t.bin(i, guard) == bin {
```
jonhoo: I _think_ you can invert this the same way the code for `put` does and just `continue` if they are _not_ equal.

domenicquirl:
I did rebase earlier, but it I somehow merged my own changes back in from remote. Rebased again.

I will look at your review comments and also more tests, but it's past midnight local time for me, so I'll call it a day for now.

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+                        validated = true;
```
jonhoo: I don't think the `validated` variable is actually necessary if you invert the clause above.

src/lib.rs
jonhoo: A comment here would be good (again, see `put`).

domenicquirl: I was thinking about this `validated` flag too, but wanted to make sure the implementation worked correctly first. Will come back to this.

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+                        // TODO: tree nodes
+                        let mut e = bin;
+                        let mut pred: Shared<'_, BinEntry<K, V>> = Shared::null();
+                        loop {
+                            // safety: either e is bin, in which case it is valid due to the above,
+                            // or e was obtained from a next pointer.
```
jonhoo: A few more words about why it's safe if it's from a next pointer would be good.

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+                            let n = unsafe { e.deref() }.as_node().unwrap();
+                            let next = n.next.load(Ordering::SeqCst, guard);
```
jonhoo: I don't think you have to load this until below the `if` (see `put`).

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+                                        .next
+                                        .store(next, Ordering::SeqCst);
+                                } else {
+                                    // or by setting the next node as the first BinEntry if there is no previous entry
+                                    t.store_bin(i, next);
+                                }
+
+                                // in either case, mark the BinEntry as garbage, since it was just removed
+                                // safety: as for val below / in put
+                                unsafe { guard.defer_destroy(e) };
```
jonhoo: I think you probably also want a `break` here.

src/lib.rs
```diff
@@ -1011,6 +1011,144 @@ where
...
+                            //    any reference to the value.
+                            //  - another thread is about to get a reference to this value.
+                            //    they execute _after_ the store_bin, and therefore do _not_ get a
+                            //    reference to the old value. there are no other ways to get to a
+                            //    value except through its Node's `value` field (which is now gone
+                            //    together with the node), so freeing the old value is fine.
+                            unsafe { guard.defer_destroy(val) };
+
+                            // safety: the lifetime of the reference is bound to the guard
+                            // supplied which means that the memory will not be modified
```
```suggestion
                            // supplied which means that the memory will not be freed
```

jonhoo:
Managed to find a bit of time to do the review! I think this looks good; just left some minor nits in the code and then I think we're pretty much ready to merge ^^

domenicquirl:
Hi! I was kind of busy the last few days, but finally managed to address your comments. Some minor additional changes to the tests were necessary to integrate the commits that were merged in the meantime (the JDK `map_check` gave me a scare when the iterator test failed, but it was only because the test depends on all tests steps being executed in an exact order).

Regarding the `jsr166` tests, I did add the one test that only tests functionality that I have implemented. However, it often refers to length and emptiness of the map, which do not yet exist in the API, so I added those as comments. The other test called `remove...` all test something which we don't have, such as catching `null` pointers and removing key-value pairs (i.e. checking against the value as well as the key). As a reminder, I did not include the latter feature due to values not currently being required to be `EQ`.

That said there are still the tests from `map_check` plus my own tests in `basic.rs`, which include removals from two threads, so I hope this is sufficient and we are ready to add `remove` 😊

src/lib.rs
```diff
@@ -1044,6 +1044,153 @@ where
...
+                                t.store_bin(i, next);
+                            }
+
+                            // in either case, mark the BinEntry as garbage, since it was just removed
+                            // safety: as for val below / in put
+                            unsafe { guard.defer_destroy(e) };
+
+                            // since the key was found and only one node exists per key, we can
+                            // break here (as an addition to the Java code)
+                            break;
```
jonhoo: The Java code also does this at
https://github.com/jonhoo/flurry/blob/8c014c349df1cbe03f98d84f7d59115899e4df7f/jsr166/src/ConcurrentHashMap.java#L1113

So I'm not sure what the comment is referring to?

jonhoo:
Yes, this looks great now! It's just the one slightly weird comment and then I think we're ready to go :tada:

jonhoo:
Oh, I'd leave the first part of the comment! Just leave out the "as opposed to Java" part.

domenicquirl:
Fixed 👍

jonhoo:
Fantastic, thanks for the change! I'll merge once the tests pass.

PR #6, opened 2020-01-22
Worked on some TODOs

GarkGarcia:
Pseudoimplemented batch operations and made it so that `num_cpus` is used to choose resize stride more intelligently.

Implemented `std::iter::FromIterator<(K, V)>` for `FlurryHashMap<K, V, _>` using the newly created `FlurryHashMap::extend` method.

The implementation of `FlurryHashMap::extend` is not efficient at all: it's intended to be a place holder for a more robust implementation. Nevertheless, now there's only one method missing an appropriate implementation (instead of two).


jonhoo:
Awesome, thank you, I'll take a look and leave some comments <3

GarkGarcia:
I just now reliazed there are some compile time erros. I'll close the PR and open a new one with the fixes.

src/lib.rs
```diff
@@ -764,8 +768,12 @@ where
         // won't be dropped while the guard remains active.
         let n = unsafe { table.deref() }.bins.len();
 
-        // TODO: use num_cpus to help determine stride
-        let stride = MIN_TRANSFER_STRIDE;
+        lazy_static! {
```
jonhoo: I'd rather not add a new dependency just for this. Can you modify this to use [`std::sync::Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) instead? I think that should be sufficient.

GarkGarcia: Sure

src/lib.rs
```diff
@@ -1073,6 +1081,18 @@ where
         let node_iter = NodeIter::new(table, guard);
         Values { node_iter, guard }
     }
+
+    /// Exntends the [`FlurryHashMap`] with the contents of `iter`.
+    // TODO: Implement this in a not as stupid way
+    pub fn extend<'g, T: IntoIterator<Item = (K, V)>>(
```
jonhoo: I think we want to also implement the [`Extend` trait](https://doc.rust-lang.org/std/iter/trait.Extend.html) for `&FlurryHashMap` now that we have this.

src/lib.rs
```diff
@@ -1073,6 +1081,18 @@ where
         let node_iter = NodeIter::new(table, guard);
         Values { node_iter, guard }
     }
+
+    /// Exntends the [`FlurryHashMap`] with the contents of `iter`.
```
```suggestion
    /// Extends the map with the contents of `iter`.
```

src/lib.rs
```diff
@@ -1073,6 +1081,18 @@ where
...
+    // TODO: Implement this in a not as stupid way
```
jonhoo: I think this may actually be the best way to implement it. The exception would be if the `Iterator` implements `ExactSizeIterator`, but that probably requires specialization. I would just remove this comment :)

Maaybe it'd be a good idea to add a `TODO` comment in the body before the loop saying
```rust
// TODO: with specialization, we could eagerly resize if adding the number of elements in an ExactSizeIterator would exceed the capacity of the map.
/// The Java version does this using the tryPresize method.
```

src/lib.rs
```diff
@@ -1093,6 +1113,17 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
     }
 }
 
+impl<K, V, S> FromIterator<(K, V)> for FlurryHashMap<K, V, S> {
+    fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) {
+        // safety: we have &mut self, so not concurrently accessed by anyone else
+        let guard = unsafe { crossbeam::epoch::unprotected() };
+        let output = Self::new();
```
jonhoo: I think I would use `Self::with_capacity` here in association with [`Iterator::size_hint`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint).

jonhoo:
You don't need to open a new PR -- just push more commits to this one :)

src/lib.rs
GarkGarcia: The problem is that since `FlurryHashMap::put` requires a guard, the `FlurryHashMap::extend` does so too. Because of that we're not able to match `Extend::extend`'s type signature (I think).

src/lib.rs
```diff
@@ -1229,5 +1232,13 @@ impl<K, V> Table<K, V> {
     }
 }
 
+#[inline]
+/// Returns the number of physical CPUs in the machine (_O(1)_).
+fn get_ncpu() -> usize {
```
jonhoo: Why is all the `unsafe` needed here?

src/lib.rs
GarkGarcia: The reason why we're able to get arround this guard thing in the implementation of `FromIterator` is that the method `FromIterator::from_iter` is garanteed to be called only at the object's initialization, thus we can use `crossbeam::epoch::unprotected()` (sinse data races are not a problem)

jonhoo: You can just create a `Guard` with `epoch::pin` inside `extend`, no?

src/lib.rs
GarkGarcia: Alright, I'll remove the comment then (and add the new one).

src/lib.rs
GarkGarcia: Makes sense. Sinse the `FlurryHashMap::extend` method does not preallocate space for the incoming entries, this should be more efficient.

By the way, couldn't we use a similar strategy for the `extend` method?

src/lib.rs
GarkGarcia: Because to avoid calling `num_cpus::get_physical` multiple times I'm storing it's return value in a `static mut`, which can only be read or mutated from unsafe code (I think).

src/lib.rs
jonhoo: Oh, yes, you're right, we don't need `ExactSizeIterator` and specialization since `size_hint` is available on all `Iterator` types! You would need to port the Java `tryPresize` method to use this in `extend` though, since you are extending an _existing_ map and therefore cannot use `with_capacity`.

src/lib.rs
jonhoo: I'm pretty sure it's safe to read without `unsafe`. Modifying it you might be right about though. Maybe make it an `AtomicUsize`, which would get rid of all the `unsafe`?

src/lib.rs
```diff
@@ -1118,7 +1119,10 @@ where
     fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> Self {
         // safety: we have &mut self, so not concurrently accessed by anyone else
         let guard = unsafe { crossbeam::epoch::unprotected() };
-        let output: FlurryHashMap<K, V, RandomState> = FlurryHashMap::new();
+        let it = iter.iter();
+        let size_hint = it.size_hint().0;
```
jonhoo: This won't actually work, since `with_capacity` panics if the requested capacity is 0, and the `min` of the `size_hint` could totally be `0`. I think maybe what we want is something like [what the standard library does for `Vec::from_iter`](https://doc.rust-lang.org/src/alloc/vec.rs.html#1982).

GarkGarcia: Fair enought, but wouldn't it be better if `FlurryHashMap::with_capacity(0)` didn't panick (return `FlurryHashMap::new()` instead)?

src/lib.rs
GarkGarcia: You're right, `epoch::pin` seems to do the trick.

src/lib.rs
GarkGarcia: Dunno, `AtomicUsize` doesn't make much sense to me, since it implies the value will be persistently modified (for me atleast). Still, I could use it if that's your decision.

src/lib.rs
GarkGarcia: Alright, I'll update the TODOs then.

src/lib.rs
jonhoo: Quite possibly -- feel free to implement that if you want to :)

jonhoo: I do like what they do in `std` for `Vec` as linked above of also calling `next()` once on the iterator to handle the special case where iterator immediately yields `None`. It's pretty smart I think.

src/lib.rs
jonhoo: I would prefer it because it removes some unnecessary `unsafe`. One day, we'll get something like [this](https://github.com/rust-lang/rfcs/pull/2788) in the standard library, and then we could move to that instead :)

jonhoo:
It looks like the main thing still failing in CI is `rustfmt`. If you run `cargo fmt` you should be good.

src/lib.rs
GarkGarcia: I didn't quite get why they did that. Could you explain a bit? I made it so that `with_capacity(0)` doesn't panick anyaway.

I'm trying to imitate what they did in `Vec::from_iter` right now.

jonhoo: They did it so that if you do `from_iter` from an iterator that doesn't yield any elements, you do not allocate anything, regardless of what the `size_hint` says. At least that's my read of it.

src/lib.rs
```diff
@@ -1093,6 +1100,46 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
...
+impl<K, V> FromIterator<(K, V)> for FlurryHashMap<K, V, RandomState>
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send,
+{
+    fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> Self {
+        let mut it = iter.into_iter();
+
+        let map = if let Some((key, value)) = it.next() {
+            let guard = crossbeam::epoch::pin();
```
jonhoo: Hmm, one concern here is that we are now using _two_ guards in order to make the map: one to do the first `put`, and one that is created inside `extend`. I think we'll probably want a (private) method
```rust
fn put_all<I: Iterator>(&self, iter: I, guard: &Guard)
```
that both the `Extend` and `FromIterator` methods call. `Extend` can make its own guard like it currently does, and here we can create a `epoch::unprotected` and use that for both the `put` and then subsequent `put_all`.

src/lib.rs
```diff
@@ -1093,6 +1100,46 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
...
+            let (lower, _) = it.size_hint();
+            let map = Self::with_capacity(lower.saturating_add(1));
+            map.put(key, value, false, &guard);
+            map
+        } else {
+            Self::new()
```
jonhoo: I think this can be a `return Self::new()`, since we know here that the `.extend` isn't necessary (since the iterator has ended).

src/lib.rs
GarkGarcia: Indeed. For now this looks like the best solution possible, but if we were to implement `tryPrecize`, we could do something similar to what's done in [`HashMap::from_iter`](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#2400-2410), moving the code in `put_all` to the `Extend` implementation and leveraging it in the `FromIterator` implementation.

What do you think?

src/lib.rs
GarkGarcia: Yep, you're right.

src/lib.rs
jonhoo: Hmm, yes, that would probably work. I think ultimately we'll likely end up with something like [`HashMap::extend`](https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L2643-L2658) (plus the `.next()` optimization from above).

We _may_ still want to special-case `FromIterator`, since it can use `with_capacity` (which is simple) rather than `tryPresize` (which is complicated), but :man_shrugging:. We definitely want to use an unprotected guard in `FromIterator` though I feel like.

src/lib.rs
```diff
@@ -1124,18 +1127,18 @@ where
 {
     fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> Self {
         let mut it = iter.into_iter();
+        let guard = crossbeam::epoch::pin();
```
jonhoo: This can use an `unprotected` guard I think

src/lib.rs
```diff
@@ -1093,6 +1106,43 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
...
+impl<K, V> FromIterator<(K, V)> for FlurryHashMap<K, V, RandomState>
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send,
+{
+    fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> Self {
+        let mut it = iter.into_iter();
+        let guard = crossbeam::epoch::pin();
+
+        let map = if let Some((key, value)) = it.next() {
```
jonhoo: You could probably avoid some indentation here with:
```rust
let first = it.next();
if first.is_none() {
    return Self::new();
}
let Some((key, value)) = first.unwrap();
// ...
```
But not sure whether it's worth it. Up to you :)

jonhoo:
I just realized that `std::collections::HashMap` also implements
```rust
Extend<(&K, &V)> where K: Copy, V: Copy
```
Up to you, but I feel like adding [that implementation](https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L2661-L2671) might also be a handy addition.

src/lib.rs
jonhoo: Ah, yes, what you did was nicer!

jonhoo:
By the way, it'd be good to add a test for `FromIterator`. I think [this test](https://github.com/jonhoo/flurry/blob/79f8d608ece6f7d26d9499b74ebfeacfa545d141/jsr166/test/ConcurrentHashMapTest.java#L603) from the Java code is the most applicable one. Since it's from a test file we don't currently have any tests for, just create a new file for it in `tests` called `jsr166.rs`!

src/lib.rs
GarkGarcia: Yep I could, but idk. I like to avoid using `unwrap` whenever possible, and since the code isn't indented that much I don't mind.

src/lib.rs
GarkGarcia: Makes sense. Let's go with it.

GarkGarcia: Anyway, `tryPrecize` should be implemented at some point. Unfortunatelly I don't feel like I have the level of understanding necessary for doing so.

src/lib.rs
GarkGarcia: Just changed it.

GarkGarcia:
> 
> 
> I just realized that `std::collections::HashMap` also implements
> 
> ```rust
> Extend<(&K, &V)> where K: Copy, V: Copy
> ```
> 
> Up to you, but I feel like adding [that implementation](https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L2661-L2671) might also be a handy addition.

Yeah, seems like a nice addition. I'll work on it.

src/lib.rs
GarkGarcia: Wow, that type is preciselly what we need! Alright then, I'll go with `AtomicUsize` then.

GarkGarcia: What ordering should I go for thought?

GarkGarcia:
> 
> 
> By the way, it'd be good to add a test for `FromIterator`. I think [this test](https://github.com/jonhoo/flurry/blob/79f8d608ece6f7d26d9499b74ebfeacfa545d141/jsr166/test/ConcurrentHashMapTest.java#L603) from the Java code is the most applicable one. Since it's from a test file we don't currently have any tests for, just create a new file for it in `tests` called `jsr166.rs`!

I'm working on it. However, I need to get hold of the map's length. Is it stored in the `count` field? Also, it would be nice to provide a `len` and a `is_empty` method to `FlurryHashMap`, I'll work on it when I figure out how to access the map's length.

src/lib.rs
```diff
@@ -1093,6 +1106,55 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
...
+impl<K, V> FromIterator<(K, V)> for FlurryHashMap<K, V, RandomState>
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send,
+{
+    fn from_iter<T: IntoIterator<Item = (K, V)>>(iter: T) -> Self {
+        let mut iter = iter.into_iter();
+
+        if let Some((key, value)) = iter.next() {
+            let guard = unsafe { crossbeam::epoch::unprotected() };
```
jonhoo: I think we should add a `// safety:` comment here

jonhoo:
> However, I need to get hold of the map's length. Is it stored in the `count` field? Also, it would be nice to provide a `len` and a `is_empty` method to `FlurryHashMap`, I'll work on it when I figure out how to access the map's length.

Yup, reading `count` should do the trick! I agree that adding those methods would be great. The Java version is in [`sumCount`](https://github.com/jonhoo/flurry/blob/79f8d608ece6f7d26d9499b74ebfeacfa545d141/jsr166/src/ConcurrentHashMap.java#L2541-L2550), but that is with the whole `CounterCell`/`LongAdder` business, which is (currently) just the `count: AtomicUsize` in the port.

src/lib.rs
jonhoo: `Relaxed` should be fine in this instance. Even _if_ a read ends up returning the original `0` value, you'll still end up with reasonable behavior (using `MIN_TRANSFER_STRIDE`).

src/lib.rs
jonhoo: That's totally fine — just leave appropriate the TODOs and they'll be a learning opportunity for someone later!

src/lib.rs
```diff
@@ -1093,6 +1106,55 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
...
 
+impl<K, V, S> Extend<(K, V)> for &FlurryHashMap<K, V, S>
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send,
+    S: BuildHasher,
+{
+    #[inline]
+    // TODO: Implement Java's `tryPresize` method to pre-allocate space for
+    // the incoming entries
```
jonhoo: Maybe add to this comment:
> `hashbrown::HashMap::extend` provides some good guidance on how to choose the presizing value based on the iterator lower bound.

GarkGarcia:
> 
> 
> > However, I need to get hold of the map's length. Is it stored in the `count` field? Also, it would be nice to provide a `len` and a `is_empty` method to `FlurryHashMap`, I'll work on it when I figure out how to access the map's length.
> 
> Yup, reading `count` should do the trick! I agree that adding those methods would be great. The Java version is in [`sumCount`](https://github.com/jonhoo/flurry/blob/79f8d608ece6f7d26d9499b74ebfeacfa545d141/jsr166/src/ConcurrentHashMap.java#L2541-L2550), but that is with the whole `CounterCell`/`LongAdder` business, which is (currently) just the `count: AtomicUsize` in the port.

Alright then. There's also another issue: at one point the Java test compares two `ConcurrentHashMap`s, so we'd have to implement `Eq` for the test to work (which I'm not sure we  want to do). How do you think going about this?

GarkGarcia:
I think the test is not building (it's not ready anyway). When I run `cargo build` in my machine everything works fine...

jonhoo:
> at one point the Java test compares two `ConcurrentHashMaps`, so we'd have to implement `Eq` for the test to work (which I'm not sure we want to do). How do you think going about this?

I think for the time being we cheat and do
```rust
// TODO: improve when `Map: Eq`
let mut fst Vec<_> = map1.iter().collect();
let mut snd Vec<_> = map2.iter().collect();
fst.sort();
snd.sort();
assert_eq!(fst, snd);
```

> I think the test is not building

You can see the compilation errors by clicking "Details" on a failed job, then clicking "1 errors / 0 warnings", and then "Bash exited with code '101'". It's awful UI, but it eventually leads you to [the log](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=405&view=logs&j=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&t=8bed322d-c384-55e1-f555-9d6311d97c43). It looks like the tests aren't building, so this would only show up if you run `cargo test`. `cargo build` does not build the tests.

GarkGarcia:
> 
> 
> > at one point the Java test compares two `ConcurrentHashMaps`, so we'd have to implement `Eq` for the test to work (which I'm not sure we want to do). How do you think going about this?
> 
> I think for the time being we cheat and do
> 
> ```rust
> // TODO: improve when `Map: Eq`
> let mut fst Vec<_> = map1.iter().collect();
> let mut snd Vec<_> = map2.iter().collect();
> fst.sort();
> snd.sort();
> assert_eq!(fst, snd);
> ```
> 
> > I think the test is not building
> 
> You can see the compilation errors by clicking "Details" on a failed job, then clicking "1 errors / 0 warnings", and then "Bash exited with code '101'". It's awful UI, but it eventually leads you to [the log](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=405&view=logs&j=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&t=8bed322d-c384-55e1-f555-9d6311d97c43). It looks like the tests aren't building, so this would only show up if you run `cargo test`. `cargo build` does not build the tests.

Fair enought.

GarkGarcia:
All done I think.

jonhoo:
@GarkGarcia Yup, I think so! Only thing left is to merge `master` and have CI pass :D

jonhoo:
I took the liberty of doing the merge -- now let's see what CI says!

jonhoo:
:tada: :heart:

GarkGarcia:
> 
> 
> 🎉 ❤️

Great! I'm glad I could help 😃

PR #7, opened 2020-01-22
Worked on some TODOs

GarkGarcia:
Pseudoimplemented batch operations and made it so that `num_cpus` is used to choose resize stride more intelligently.

Implemented `std::iter::FromIterator<(K, V)>` for `FlurryHashMap<K, V, _>` using the newly created `FlurryHashMap::extend` method.

The implementation of `FlurryHashMap::extend` is not efficient at all: it's intended to be a place holder for a more robust implementation. Nevertheless, now there's only one method missing an appropriate implementation (instead of two).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,5 @@ maintenance = { status = "experimental" }
 crossbeam = "0.7"
 parking_lot = "0.10"
 rand = "0.7"
-lazy_static = "1.4.0"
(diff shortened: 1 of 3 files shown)


jonhoo:
You don't need to open a new PR -- just push more commits to the old one :)

jonhoo:
The other PR is #6 for anyone keeping track.

Issue #10, opened 2020-01-23
Implement "standard" map/collection methods


jonhoo:
This issue tracks the implementation status of various "standard" methods on collections in general and maps in particular that should be added to flurry. Checked boxes have already been "claimed".

 - [ ] [`drain`] (**hard**) and `IntoIterator` (**easy**) for both owned and `&` (see #33)

If you want to pick up any of these, leave a comment here and/or open a draft PR!

  [`drain`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.drain

Stupremee:
I would like to implement Eq, Index and clear.

jonhoo:
@Stupremee How about you do `Eq` and `Index` first, and then we do `clear` in a separate PR afterwards?

danielSanchezQ:
I think I can try to implement the `Clone` one . Is it ok?

jonhoo:
Absolutely, it's all yours!

Stupremee:
I will also implement Debug since Index doesnt work atm

danielSanchezQ:
I think I can do `with_hasher` and `with_capacity_and_hasher` also. But I see them as checked already.

jonhoo:
@danielSanchezQ Those are already (accidentally) being added in #24

soruh:
I'd like to take a crack at implementing `reserve`.

jonhoo:
@soruh all yours!

danielSanchezQ:
I can give a try to the `retain` method 😄

soruh:
I took a quick look at `drain` and `into_iter` in #33, but I'm not sure how correct & efficient what I wrote actually is.

soruh:
@jonhoo I'd appreciate it if you could mark the `drain` and `IntoIter` impls as open (maybe with a link to my existing work); I don't really have the time to look into this further at the moment (I'm still open for questions about my initial implementation #33). I'll write another comment if I have the time to work on this and no one already has.

jonhoo:
I'm replacing this with https://github.com/jonhoo/flurry/issues/62 now that there's just one item left.

PR #18, opened 2020-01-24
Implemented `get_key_value`

jonathanGB:
Referring to #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -404,11 +404,10 @@ where
     }
 
     /// Returns the key-value pair corresponding to the supplied key.
-    /// 
+    ///
     /// Returns `None` if this map contains no mapping for `key`.
     pub fn get_key_value<'g>(&'g self, key: &'g K, guard: &'g Guard) -> Option<(&'g K, &'g V)> {
-        self.get(key, guard)
-            .map(|v| (key, v))
+        self.get(key, guard).map(|v| (key, v))
     }
 
     fn init_table<'g>(&self, guard: &'g Guard) -> Shared<'g, Table<K, V>> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -468,11 +468,11 @@ where
     }
 
     /// Returns the key-value pair corresponding to `key`.
-    /// 
+    ///
     /// Returns `None` if this map contains no mapping for `key`.
     ///
     /// The supplied `key` may be any borrowed form of the
-    /// map's key type, but `Hash` and `Eq` on the borrowed form 
+    /// map's key type, but `Hash` and `Eq` on the borrowed form
     /// must match those for the key type.
     pub fn get_key_value<'g, Q>(&'g self, key: &Q, guard: &'g Guard) -> Option<(&'g K, &'g V)>
     where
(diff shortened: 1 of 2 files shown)


jonhoo:
Excellent, thanks, let me take a look!

src/lib.rs
```diff
@@ -403,6 +403,13 @@ where
         self.get(key, guard).map(then)
     }
 
+    /// Returns the key-value pair corresponding to the supplied key.
+    ///
+    /// Returns `None` if this map contains no mapping for `key`.
+    pub fn get_key_value<'g>(&'g self, key: &'g K, guard: &'g Guard) -> Option<(&'g K, &'g V)> {
```
jonhoo: I don't think you need the `'g` on the `self` reference here. As long as the guard lives, the pointers will remain valid, even if `self` completely goes away.

jonhoo: In fact, I don't think you need it on `key` either, since we return the key we read from the map.

jonhoo: Actually, damn, I take that back. I think there's a bug in `insert`. The `'g` _does_ have to be associated with the lifetime of `self` for [this](https://github.com/jonhoo/flurry/blob/6d56c58026f6edd0d9828ce48b3cbb6aecd38973/src/lib.rs#L1227) to be safe. It shouldn't be needed on `key` though!

src/lib.rs
```diff
@@ -403,6 +403,13 @@ where
         self.get(key, guard).map(then)
     }
 
+    /// Returns the key-value pair corresponding to the supplied key.
+    ///
+    /// Returns `None` if this map contains no mapping for `key`.
```
jonhoo: It'd be good to match the documentation and bounds on [`HashMap::get_key_value`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get_key_value) here.

src/lib.rs
jonhoo: Ah, I see, below you _are_ returning the `key` that is passed in. That is probably not what we want to do. Instead, we'll want an internal version of `get` that produces a `&Node`, and then have `get` call that and extract the key, and this have call that and extract both the key and the value.

cuviper: Following #25, this should probably use a borrowed `Q` type for the key.

jonathanGB: Yep, I had started changing everything to `&Q` but I see it's done :P

src/lib.rs
```diff
@@ -432,7 +424,24 @@ where
         // next epoch after it is removed. since it wasn't removed, and the epoch was pinned, that
         // cannot be until after we drop our guard.
         let node = unsafe { node.deref() };
-        let node = node.as_node().unwrap();
+        node.as_node()
```
jonhoo: I think this should actually be
```rust
Some(node.as_node().expect("BinEntry::find should always return Node"))
```

jonathanGB: I had the callers of `get_node` use the `?` operator, but I'm fine with that as well

jonathanGB: Oh wait, no I kept it as an option because some sanity check in the method might return `None` (see lines 388 and 395).

jonhoo: Ah, no, this isn't about the return type of _this_ method (notice that we are re-wrapping in `Some`). It is about having an assertion that checks that `BinEntry::find` _indeed_ always returns a `Node`.

jonathanGB: (forget all I said)

Issue #17, opened 2020-01-23
Add a Set type matching the std HashSet API


cuviper:
Just like in `std`, it can start with the simple definition wrapping a map:

```rust
pub struct FlurryHashSet<T, S = RandomState> {
    map: FlurryHashMap<T, (), S>,
}
```

jonhoo:
The Java code's [`KeySetView`](https://github.com/jonhoo/flurry/blob/8c014c349df1cbe03f98d84f7d59115899e4df7f/jsr166/src/ConcurrentHashMap.java#L4563) may also be of interest.

jonhoo:
We now have a basic `HashSet` type, but it'd be great to get something along the lines of `HashMapRef` for `HashSet` too!

GarkGarcia:
> We now have a basic `HashSet` type, but it'd be great to get something along the lines of `HashMapRef` for `HashSet` too!

Is that a case of simply wrapping `HashMapRef<T, ()>` in a newtype? I could work on that.

twe4ked:
I _think_ you'd want to wrap `HashSet` around with `HashSetRef` similar to how `HashMap` is wrapped by `HashMapRef`.

GarkGarcia:
> I _think_ you'd want to wrap `HashSet` with `HashSetRef` similar to how `HashMap` is wrapped by `HashMapRef`.

That's interesting. I'm not quite sure which way would be the most efficient memory wise. @jonhoo @cuviper opinions?

GarkGarcia:
We'd be wrapping a wrapper either way...
I think wrapping `HashMap` directly should make the type's layout flatter, but I'm not sure.

jonhoo:
Ah, no, I was imagining we just do the same trick with `HashSetRef` as with `HashMapRef`, not that we wrap `HashMapRef` to make `HashSetRef`!

GarkGarcia:
> Ah, no, I was imagining we just do the same trick with `HashSetRef` as with `HashMapRef`, not that we wrap `HashMapRef` to make `HashSetRef`!

Makes sense.

GarkGarcia:
@jonhoo What about the _wrapping `HashMap<T, ()>` directly_ proposal?

jonhoo:
I think we should wrap `HashSet`, because `HashSet` actually has different _methods_ than `HashMap`, and having the wrapper _also_ do the translation from set-like methods to map-like methods (which `HashSet` is already doing) seems like unfortunate duplication.

[...]

PR #19, opened 2020-01-24
Add basic HashSet implementation

twe4ked:
Add a very basic implementation of `HashSet` as mentioned in #17.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/19)


src/set/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+            map: FlurryHashMap::<T, ()>::new(),
+        }
+    }
+
+    /// Adds a value to the set.
+    ///
+    /// If the set did not have this value present, true is returned.
+    ///
+    /// If the set did have this value present, false is returned.
+    pub fn insert(&self, value: T) -> bool {
```
twe4ked: I'm a little unclear as to the semantics of when a guard needs to be passed in vs constructed inside the public API. AFAICT it's to do with the lifetime of the returned value, and since we're not returning anything here I think this is okay?

jonhoo: That is exactly correct -- the guard is needed to ensure that the value a reference we return points to lives for as long as the reference does. Since we here do not return any reference, we do not have to take a guard in either, since it's fine for the lifetime of `old` to not outlast the current function (which is where `guard` will be dropped).

jonhoo:
This looks great! Three comments:

 - I'd like to see the docs slightly expanded to be more similar to those in the [standard library](https://doc.rust-lang.org/std/collections/struct.HashSet.html). You don't need to go into _quite_ as much detail, but some of it would be useful.
 - There are some additional methods that would be good to forward to the inner map. Off the top of my head, `.remove()` and `.iter()` (which'll forward to `.keys()`) are the most obvious candidates.
 - We should port some of the Java tests that are related to sets. From a quick glance, those seem to be [these](https://github.com/jonhoo/flurry/blob/6d56c58026f6edd0d9828ce48b3cbb6aecd38973/jsr166/test/ConcurrentHashMapTest.java#L323) and [these](https://hg.openjdk.java.net/jdk/jdk13/file/0368f3a073a9/test/jdk/java/util/concurrent/ConcurrentHashMap/MapCheck.java#l236). They are _technically_ for a set view _of_ a map, but should be straightforward enough to adapt to this a standalone set I think.

twe4ked:
I haven't added any additional tests yet (aside from the doc tests). I haven't had a chance to take a look at the examples you linked but I'm not sure how much they're needed yet as this is only using the public API which is already tested.

I'll still take a look at adding some of the ones you linked.

twe4ked:
>  codecov/patch — 57.14% of diff hit (target 82.7%) 

Oh interesting, it doesn't count doc tests.

jonhoo:
I think the reason to add tests to this is more in case we decide to make more optimizations specifically to the set version later.

Not sure what you mean by doctests not being covered? They [should be](https://github.com/xd009642/tarpaulin#features). There was [an issue](https://github.com/xd009642/tarpaulin/issues/13) about it in the past, but that appears to have been solved.

src/set/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// ```
+    /// use flurry::set::FlurryHashSet;
+    ///
+    /// let set = FlurryHashSet::new();
+    /// set.insert(2);
+    ///
+    /// assert!(set.contains(&2));
+    /// assert!(!set.contains(&1));
+    /// ```
+    pub fn contains(&self, value: &T) -> bool {
```
cuviper: I'd suggest taking a borrowed `Q` type here, as mentioned in #25.

src/set/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// use flurry::set::FlurryHashSet;
+    ///
+    /// let set = FlurryHashSet::new();
+    /// set.insert(2);
+    ///
+    /// assert_eq!(set.remove(&2), true);
+    /// assert!(!set.contains(&2));
+    /// assert_eq!(set.remove(&2), false);
+    /// ```
+    pub fn remove(&self, value: &T) -> bool {
```
cuviper: This can take a borrowed `Q` type too.

jonhoo:
Just a heads up that I moved a bunch of code around (though it is mostly unchanged) and renamed `FlurryHashMap` to just `HashMap`. Shouldn't be too hard to merge with, but just so you are prepared for git complaining :sweat_smile:

twe4ked:
> Not sure what you mean by doctests not being covered? They should be. There was an issue about it in the past, but that appears to have been solved.

I guess I got confused by the diff code cov dropping. I might have been the changes I rebased in. I still don't the coverage isn't 100%. Should be fixed when I add in the other tests.

> Just a heads up that I moved a bunch of code around

Thanks for the heads up!

cuviper:
When you get back to this, it would be nice to also have a `HashSetRef` like #45 did for maps. That could be a separate PR though.

twe4ked:
> This branch is out-of-date with the base branch

Haha, I just rebased this branch!

All good, taking a look at the tests now :)

src/set/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+#[derive(Debug)]
+pub struct HashSet<T: 'static, S = crate::DefaultHashBuilder>
+where
+    T: Sync + Send + Clone + Hash + Eq,
+    S: std::hash::BuildHasher,
+{
+    map: HashMap<T, (), S>,
+}
+
+impl<T> HashSet<T, crate::DefaultHashBuilder>
```
cuviper: `DefaultHashBuilder` is correct for `fn new`, but the other methods should allow a generic `S: BuildHasher`. Consider the others too: `with_capacity` using the default, `with_hasher` and `with_capacity_and_hasher` using `S`.

twe4ked:
> All good, taking a look at the tests now :)

I had a look at the relevant tests from the Java codebase and honestly it looked like everything is covered by the doc tests. They were all pretty simple tests that didn't seem hugely relevant to how this is implemented here.

I'm happy to “finish” this PR off by adding the missing constructor methods and getting it back up to date with `master` but I don't think I'll have time for much further than that. Also happy to close as I haven't done anything groundbreaking here and it should be easy for someone else to pick up if they want to :)

`HashSetRef` also sounds like a great idea but, as mentioned, probably shouldn't block this PR being merged.

Thoughts?

jonhoo:
If you bring it up to speed with master, I'd be happy to merge without a `HashSetRef` :)

Sticking to just the doc tests seems fine by me!

twe4ked:
Tests are failing with:

```
/bin/bash --noprofile --norc /Users/runner/runners/2.164.8/work/_temp/b4288b92-8bed-41ec-ad74-9fcb74730d03.sh
    Updating git repository `https://github.com/cynecx/crossbeam.git`
    Updating crates.io index
error: failed to select a version for the requirement `crossbeam-epoch = "^0.9"`
  candidate versions found which didn't match: 0.8.0, 0.7.2, 0.7.1, ...
  location searched: crates.io index
```

Looks like https://github.com/crossbeam-rs/crossbeam/pull/458 has been merged which may have broken something?

The [branch](https://github.com/cynecx/crossbeam/commits/fix-unsoundness) is still around though, so not sure why it can't use it.

jonhoo:
Ah, sorry, should be fixed on `master` now!

twe4ked:
The doc tests still don't seem to be included in the coverage tests:

https://codecov.io/gh/jonhoo/flurry/compare/7c1523de91e6e9615c8e30787c43a76a1c3e6f39...20a5f863e1281e50b30091a288cdf8e9ebecb3d8/diff#D3-77

jonhoo:
Ah, yes. I have a fix pending for that in https://github.com/jonhoo/flurry/pull/63.

jonhoo:
`master` should now include doctests in coverage

twe4ked:
```
/bin/bash --noprofile --norc /home/vsts/work/_temp/ea0e731b-71c1-4fb0-bdd9-a3623a442bb5.sh
error: component 'miri' for target 'x86_64-unknown-linux-gnu' is unavailable for download for channel nightly
Sometimes not all components are available in any given nightly.
```

I think miri might be unavailable on nightly at the moment?

https://rust-lang.github.io/rustup-components-history/

jonhoo:
Yup, it is — see also https://github.com/jonhoo/flurry/pull/61#issuecomment-589820944. Should hopefully be fixed soon-ish.

twe4ked:
It finally passed! 🎉

twe4ked:
I've actioned your feedback except for adding `HashSet::take` and `HashSet::replace`. I had a quick look into them and it wasn't immediately clear to me how to go about implementing it and I don't have time to dig into it right now.

Looks like miri is unavailable again.

jonhoo:
Thanks! I'm going to go ahead and merge this, and then I'll do some tweaks after the fact since you're short on time :)

jonhoo:
See https://github.com/jonhoo/flurry/pull/68

PR #21, opened 2020-01-24
Implement Eq and Debug traits

Stupremee:
This pr implements the `PartialEq`, `Eq` and the `Debug` traits.
These additions were requested in #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1253,6 +1253,28 @@ where
     }
 }
 
+impl<K, V, S> PartialEq for FlurryHashMap<K, V, S>
+where
+    K: Sync + Send + Clone + Eq + Hash,
+    V: Sync + Send + PartialEq,
+    S: BuildHasher,
+{
+    fn eq(&self, other: &Self) -> bool {
+        let guard = epoch::pin();
+
+        self.iter(&guard)
+            .all(|(key, value)| other.get(key, &guard).map_or(false, |v| *value == *v))
+    }
+}
+
+impl<K, V, S> Eq for FlurryHashMap<K, V, S>
+where
+    K: Sync + Send + Clone + Eq + Hash,
+    V: Sync + Send + PartialEq,
+    S: BuildHasher,
+{
+}
+
 impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
     fn drop(&mut self) {
         // safety: we have &mut self _and_ all references we have returned are bound to the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -200,16 +200,12 @@ use node::*;
 use crossbeam::epoch::{Atomic, Guard, Owned, Shared};
 use std::collections::hash_map::RandomState;
 use std::hash::{BuildHasher, Hash};
-<<<<<<< HEAD
 use std::iter::FromIterator;
+use std::ops::Index;
 use std::sync::{
     atomic::{AtomicIsize, AtomicUsize, Ordering},
     Once,
 };
-=======
-use std::ops::Index;
-use std::sync::atomic::{AtomicIsize, AtomicUsize, Ordering};
->>>>>>> Change V to Eq in Eq impl, check equality in tests both ways, add values to both maps in different size test
 
 /// The largest possible table capacity.  This value must be
 /// exactly 1<<30 to stay within Java array allocation and indexing
@@ -1265,8 +1261,11 @@ where
     S: BuildHasher,
 {
     fn eq(&self, other: &Self) -> bool {
-        let guard = epoch::pin();
+        if self.len() != other.len() {
+            return false;
+        }
 
+        let guard = epoch::pin();
         self.iter(&guard)
             .all(|(key, value)| other.get(key, &guard).map_or(false, |v| *value == *v))
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -199,6 +199,7 @@ use node::*;
 
 use crossbeam::epoch::{Atomic, Guard, Owned, Shared};
 use std::collections::hash_map::RandomState;
+use std::fmt::{self, Debug, Formatter};
 use std::hash::{BuildHasher, Hash};
 use std::iter::FromIterator;
 use std::sync::{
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1227,6 +1227,28 @@ where
...
 
+impl<K, V, S> PartialEq for FlurryHashMap<K, V, S>
+where
+    K: Sync + Send + Clone + Eq + Hash,
+    V: Sync + Send + PartialEq,
+    S: BuildHasher,
+{
+    fn eq(&self, other: &Self) -> bool {
+        let guard = epoch::pin();
+
```
jonhoo: I believe you also have to check `self.len() == other.len()` here

src/lib.rs
```diff
@@ -1227,6 +1227,28 @@ where
...
+        self.iter(&guard)
+            .all(|(key, value)| other.get(key, &guard).map_or(false, |v| *value == *v))
+    }
+}
+
+impl<K, V, S> Eq for FlurryHashMap<K, V, S>
+where
+    K: Sync + Send + Clone + Eq + Hash,
+    V: Sync + Send + PartialEq,
```
jonhoo: We are probably only `Eq` if `V` is also `Eq`.

tests/basic.rs
```diff
@@ -162,6 +162,51 @@ fn current_kv_dropped() {
...
+    let map2 = FlurryHashMap::<usize, usize>::new();
+    {
+        let guard = epoch::pin();
+        map1.insert(1, 0, &guard);
+        map2.insert(1, 1, &guard);
+    }
+
+    assert_ne!(map1, map2);
+}
+
```
jonhoo: We should add a test where the two maps are of different sizes.

We should also have all of these test for equality both ways:
```rust
assert_eq!(map1, map2);
assert_eq!(map2, map1);
```

src/lib.rs
Stupremee: Yep, I know, but there's currently no `len` method. 
I would need to  create two iterators for both maps to compare the length atm

jonhoo: The `len` method is coming in https://github.com/jonhoo/flurry/pull/6, so I guess we should just wait for that. Or just copy that part of the change into here :)

jonhoo:
I think it would be good to call out in the documentation (probably under "Consistency") the fact that checking for map equality might yield weird results if the map is being concurrently modified!

src/lib.rs
Stupremee: Okay, I will wait for #6 until this PR is ready, otherwise I will copy the part

jonhoo: Should just be a couple of minutes and then I'll merge it :)

Stupremee:
Tests are failing because the len check is missing in PartialEq trait

src/lib.rs
```diff
@@ -1250,6 +1250,20 @@ where
...
+where
+    K: Sync + Send + Clone + Eq + Hash,
+    V: Sync + Send,
+    S: BuildHasher,
+{
+    type Output = V;
+
+    fn index(&self, idx: K) -> &Self::Output {
+        let guard = epoch::pin();
+        self.get(&idx, &guard).expect("no entry found for key")
```
Stupremee: This line fails to compile with `cannot return value referencing local variable 'guard'`, but I have no idea how I can fix this.

src/lib.rs
jonhoo: All right, it's landed on `master`!

src/lib.rs
jonhoo: Ah, yes, the return value from `get` is tied to the lifetime of the guard. The guard is what ensures that the reference we return does not get garbage collected while we still have a reference to it! Since you construct the guard inside the function, it gets dropped when the function returns, but you are trying to return the reference nonetheless.

This is definitely a little annoying. The way we get around this for our other methods is that we take the guard as _input_ to the function, but that won't quite work here due to the trait forcing a particular method signature. I wonder if you could do:
```rust
impl<'a, K, V, S> Index for (&'a FlurryHashMap<K, V, S>, &'a Guard) {
```
It's not super pretty, but I _think_ it would let people write code like:
```rust
let map = ...;
let guard = epoch::pin();
let v = &(&map, &guard)[key];
```
Although on second thought, maybe not, since the lifetime of the returned reference would be tied to the lifetime of the tuple rather than the `'a`... Not sure how to go about this actually. Maybe try it and see what happens? You could try
```rust
type Output = &'a V;
```
In which case the return value would be of type `&&'a V`. It's a little awkward, but not _terrible_.

I think the best path forward is to try the above and then see what _using_ the `Index` trait feels like. If it is too cumbersome we can just leave it out.

Stupremee: Okay I will try it, thanks

Stupremee:
I recreated your idea [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e9967935bd1038f32ea55bb3731bacd0) and it doesn't work.

Stupremee:
I will just leave it as it is and we can implement it later if possible.
Unless you find a solution

jonhoo:
Ah, that is too bad. Well, since the current `Index` impl doesn't work either, I think we should then just remove the impl of `Index` for the time being :)

Stupremee:
Okay. I will just add Debug impl too

tests/basic.rs
```diff
@@ -7,6 +7,18 @@ fn new() {
...
+fn debug_format() {
+    let map = FlurryHashMap::<usize, usize>::new();
+    {
+        let g = epoch::pin();
+        map.insert(1, 0, &g);
+        map.insert(2, 1, &g);
+        map.insert(3, 2, &g);
+        map.insert(4, 3, &g);
+    }
+    assert_eq!("{3: 2, 1: 0, 2: 1, 4: 3}", format!("{:?}", map));
```
jonhoo: Hmm, this assertion will probably fail since iteration order is random..

Stupremee: Yep, already removed it

jonhoo:
Excellent, thank you!

PR #22, opened 2020-01-24
FlurryHashMap Clone

danielSanchezQ:
Implemented Clone for FlurryHashMap
Implemented Clone for Table
Added simple clone map test by checking values on preloaded map and cloned one

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1251,9 +1251,30 @@ impl<K, V, S> Drop for FlurryHashMap<K, V, S> {
     }
 }
 
+impl<K, V> Clone for FlurryHashMap<K, V, RandomState> {
+    fn clone(&self) -> Self {
+        Self {
+            table: self.table.clone(),
+            next_table: self.next_table.clone(),
+            transfer_index: AtomicIsize::new(self.transfer_index.load(Ordering::SeqCst)),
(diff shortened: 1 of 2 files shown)


danielSanchezQ:
I'm working on updating the branch with the latest changes

@jonhoo , It was simple I think but I have the feeling I'm missing something here. Should the Clone implementation be concurrent safer somehow? Thanks

jonhoo:
Hmm, I think that's probably not a safe way to do it, because you will be aliasing all of the keys, values, and nodes. They place to start is probably to do the same thing as the [Java `Map` constructor](https://github.com/jonhoo/flurry/blob/f6cdb55aca1d15e863f3790cad9dfb3095ec69f7/jsr166/src/ConcurrentHashMap.java#L821), which really just allocates a new map (with the necessary capacity) and then inserts all the elements from the old map one at a time. In Rust you will have to clone each key/value pair you iterate over before inserting into the new map, but that should be fine!

danielSanchezQ:
That makes completely sense. This should have been a draft instead, sorry for that. I'll just close this one a open a proper one next time.

jonhoo:
Don't worry about it -- feel free to just re-open this one and continue using it :)

PR #23, opened 2020-01-24
HashMap cloning

danielSanchezQ:


src/lib.rs
```diff
@@ -1353,6 +1353,23 @@ where
     }
 }
 
+impl<K, V> Clone for FlurryHashMap<K, V, RandomState>
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send + Clone,
+{
+    fn clone(&self) -> Self {
+        let cloned_map = FlurryHashMap::with_capacity(self.count.load(Ordering::SeqCst));
```
jonhoo: You should be able to use `self.len()` now that #6 has landed!

src/lib.rs
```diff
@@ -1353,6 +1353,23 @@ where
...
+        {
+            let guard = epoch::pin();
+            for (k, v) in self.iter(&guard) {
+                cloned_map.put(k.clone(), v.clone(), false, &epoch::pin());
```
```suggestion
                cloned_map.insert(k.clone(), v.clone(), &guard);
```

src/lib.rs
```diff
@@ -1353,6 +1353,23 @@ where
     }
 }
 
+impl<K, V> Clone for FlurryHashMap<K, V, RandomState>
```
jonhoo: We should implement this for any `S` too, not just `RandomState`

src/lib.rs
danielSanchezQ: Awesome, I was looking for it and couldn't find it. It is much better, thanks!

src/lib.rs
```diff
@@ -1353,6 +1353,23 @@ where
...
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send + Clone,
+{
+    fn clone(&self) -> Self {
+        let cloned_map = FlurryHashMap::with_capacity(self.count.load(Ordering::SeqCst));
+        {
+            let guard = epoch::pin();
+            for (k, v) in self.iter(&guard) {
```
jonhoo: Let us add a note under "consistency" in the top-level documentation that `Clone` may not produce a "perfect" clone if the underlying map is being concurrently modified.

src/lib.rs
danielSanchezQ: Absolutely right.

tests/basic.rs
```diff
@@ -185,3 +185,18 @@ fn drop_value() {
...
     assert_eq!(Arc::strong_count(&dropped2), 1);
 }
+
+#[test]
+// Test that same values exists in both maps (original and cloned)
+fn clone_map() {
+    let map = FlurryHashMap::<&'static str, u32>::new();
+    map.insert("FooKey", 0, &epoch::pin());
+    map.insert("BarKey", 10, &epoch::pin());
+    let cloned_map = map.clone();
```
jonhoo: Maybe add a comment here saying
```rust
// TODO: use assert_eq once #21 lands
```

tests/basic.rs
```diff
@@ -185,3 +185,18 @@ fn drop_value() {
     // Second NotifyOnDrop was dropped when the map was dropped
     assert_eq!(Arc::strong_count(&dropped2), 1);
 }
+
+#[test]
+// Test that same values exists in both maps (original and cloned)
+fn clone_map() {
```
jonhoo: It might be good to take a look through the various Java tests and see if any of those test the constructor that copies an existing map.

src/lib.rs
danielSanchezQ: That was one of my concerns. How could we do it? I guess some `RWLock`, but is it worth?

jonhoo: Ah, no, I think we specifically _want_ `clone` to be racy. We just need to make sure to call it out in the documentation, just like the Java version does :)

tests/basic.rs
danielSanchezQ: This is what they are testing. I added the part in where they test they are not mapping the same tables.

``` java
    /**
     * ConcurrentHashMap(map) creates a new map with the same mappings
     * as the given map
     */
    public void testConstructor5() {
        ConcurrentHashMap map1 = map5();
        ConcurrentHashMap map2 = new ConcurrentHashMap(map5());
        assertTrue(map2.equals(map1));
        map2.put(one, "F");
        assertFalse(map2.equals(map1));
    }
```

jonhoo: Yeah, changing one and then making sure that the cloned-from map does not also change seems like a good thing to test.

src/lib.rs
```diff
@@ -1353,6 +1356,32 @@ where
...
 
+impl<K, V, S> Clone for FlurryHashMap<K, V, S>
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send + Clone,
+    S: BuildHasher + Clone,
+{
+    fn clone(&self) -> FlurryHashMap<K, V, S> {
+        // TODO: This should be created with the FlurryHashMap::with_capacity_and_hasher once available
+        let cloned_map = Self {
```
danielSanchezQ: I had some trouble with the type constrains here. Since `new` and `with_capacity` are implemented for `FlurryHashMap<K, V, RandomState>`. I think (and added a comment) that once the `with_capacity_and_hasher` is implemented it should be use here.

src/lib.rs
```diff
@@ -40,6 +40,9 @@
 //! updates in other threads. Otherwise the results of these methods reflect transient states that
 //! may be adequate for monitoring or estimation purposes, but not for program control.
 //!
+//! [`Clone`](FlurryHashMap::clone) may not produce a "perfect" clone if the underlying map
```
jonhoo: I would fold this into the previous paragraph by saying:
> Similarly, `Clone` may not ...

src/lib.rs
```diff
@@ -1353,17 +1356,26 @@ where
...
+impl<K, V, S> Clone for FlurryHashMap<K, V, S>
 where
     K: Sync + Send + Clone + Hash + Eq,
     V: Sync + Send + Clone,
+    S: BuildHasher + Clone,
 {
-    fn clone(&self) -> Self {
-        let cloned_map = FlurryHashMap::with_capacity(self.count.load(Ordering::SeqCst));
+    fn clone(&self) -> FlurryHashMap<K, V, S> {
+        // TODO: This should be created with the FlurryHashMap::with_capacity_and_hasher once available
```
jonhoo: We'll have that once https://github.com/jonhoo/flurry/pull/24 lands!

src/lib.rs
jonhoo: Yeah, let's hold off on this until #24 lands, since that'll give us the necessary method.

src/lib.rs
danielSanchezQ: Awesome, we can wait till then I think.

tests/basic.rs
```diff
@@ -185,3 +185,33 @@ fn drop_value() {
...
     assert_eq!(Arc::strong_count(&dropped2), 1);
 }
+
+#[test]
+fn clone_map_empty() {
+    let map = FlurryHashMap::<&'static str, u32>::new();
+    let cloned_map = map.clone();
+    assert_eq!(map.len(), cloned_map.len());
+    assert_eq!(cloned_map.len(), 0);
+    // TODO: use assert_eq once #21 lands
```
jonhoo: #21 has landed!

tests/basic.rs
```diff
@@ -185,3 +185,33 @@ fn drop_value() {
...
+
+#[test]
+// Test that same values exists in both maps (original and cloned)
+fn clone_map_filled() {
+    let map = FlurryHashMap::<&'static str, u32>::new();
+    map.insert("FooKey", 0, &epoch::pin());
+    map.insert("BarKey", 10, &epoch::pin());
+    let cloned_map = map.clone();
+    assert_eq!(map.len(), cloned_map.len());
+    // TODO: use assert_eq once #21 lands
```
jonhoo: #21 has landed!

src/lib.rs
```diff
@@ -40,7 +40,7 @@
 //! updates in other threads. Otherwise the results of these methods reflect transient states that
 //! may be adequate for monitoring or estimation purposes, but not for program control.
 //!
-//! [`Clone`](FlurryHashMap::clone) may not produce a "perfect" clone if the underlying map
+//! Similarly, [`Clone`](FlurryHashMap::clone) may not produce a "perfect" clone if the underlying map
```
jonhoo: And, sorry for the nit, let's remove the empty line above this so it falls into the same paragraph :sweat_smile:

src/lib.rs
jonhoo: #24 just landed :tada:

jonhoo:
@danielSanchezQ I think this is ready to merge once the two comments above are resolved! :tada:

jonhoo:
Also, as an aside, you do not have to force-push every time. It is actually easier to review if you just push as new commits, and then we can always squash at the end :)

danielSanchezQ:
https://github.com/jonhoo/flurry/pull/23#issuecomment-578276868

It is just the way I update the branch by rebasing instead of merging (the way Im used to work). Sorry 😅

jonhoo:
@danielSanchezQ hehe, yup, I know, and it's nice when trying to keep your own history neat and tidy, but it's not so great for review, mostly because GitHub isn't great at managing the review process when all the commits change (which is what happens when you rebase).

jonhoo:
Awesome, thank you! :heart:

danielSanchezQ:
Thank you too. I'll stick around helping :)

PR #24, opened 2020-01-24
Fixed bug in `Node#find`

jonathanGB:
It turns out that `find` would always return the first entry in a bucket, as longs as the desired entry was found somewhere in the bucket. This can be replicated by reverting the change in `lib/node.rs`.

At the same time, I added two new constructors of `FlurryHashMap`, which can take a custom **Hasher**: the architecture allowed other hashers to be used, but we couldn't actually build any `Flurry` instances beforehand.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -299,55 +299,50 @@ where
 {
     /// Creates a new, empty map with the default initial table size (16).
     pub fn new() -> Self {
-        Self {
-            table: Atomic::null(),
-            next_table: Atomic::null(),
-            transfer_index: AtomicIsize::new(0),
-            count: AtomicUsize::new(0),
-            size_ctl: AtomicIsize::new(0),
-            build_hasher: RandomState::new(),
-        }
+        Self::with_hasher(RandomState::new())
     }
 
     /// Creates a new, empty map with an initial table size accommodating the specified number of
     /// elements without the need to dynamically resize.
     pub fn with_capacity(n: usize) -> Self {
-        if n == 0 {
-            return Self::new();
-        }
-
-        let mut m = Self::new();
-        let size = (1.0 + (n as f64) / LOAD_FACTOR) as usize;
-        // NOTE: tableSizeFor in Java
-        let cap = std::cmp::min(MAXIMUM_CAPACITY, size.next_power_of_two());
-        m.size_ctl = AtomicIsize::new(cap as isize);
-        m
+        Self::with_capacity_and_hasher(RandomState::new(), n)
     }
 }
 
 impl<K, V, S: BuildHasher> FlurryHashMap<K, V, S> {
-    /// Creates a new, empty map with the default initial table size (16).
-    /// It uses the given hasher, rather than the default `RandomState`.
-    pub fn new_with_hasher(hasher: S) -> Self {
+    /// Creates an empty HashMap which will use the given hash builder to hash keys.
+    ///
+    /// The created map has the default initial capacity.
+    ///
+    /// Warning: hash_builder is normally randomly generated, and is designed to
+    /// allow HashMaps to be resistant to attacks that cause many collisions and
+    /// very poor performance. Setting it manually using this
+    /// function can expose a DoS attack vector.
+    pub fn with_hasher(hash_builder: S) -> Self {
         Self {
             table: Atomic::null(),
             next_table: Atomic::null(),
             transfer_index: AtomicIsize::new(0),
             count: AtomicUsize::new(0),
             size_ctl: AtomicIsize::new(0),
-            build_hasher: hasher,
+            build_hasher: hash_builder,
         }
     }
 
-    /// Creates a new, empty map with an initial table size accommodating the specified number of
-    /// elements without the need to dynamically resize.
-    /// It uses the given hasher, rather than the default `RandomState`.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -325,14 +325,44 @@ where
     }
 }
 
+impl<K, V, S: BuildHasher> FlurryHashMap<K, V, S> {
+    /// Creates a new, empty map with the default initial table size (16).
+    /// It uses the given hasher, rather than the default `RandomState`.
+    pub fn new_with_hasher(hasher: S) -> Self {
```
jonhoo: This method is traditionally called [`with_hasher`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_hasher). It'd be good to match the documentation to that of the standard library for the same method!

src/lib.rs
```diff
@@ -325,14 +325,44 @@ where
...
+            count: AtomicUsize::new(0),
+            size_ctl: AtomicIsize::new(0),
+            build_hasher: hasher,
+        }
+    }
+
+    /// Creates a new, empty map with an initial table size accommodating the specified number of
+    /// elements without the need to dynamically resize.
+    /// It uses the given hasher, rather than the default `RandomState`.
+    pub fn with_hasher_and_capacity(hasher: S, n: usize) -> Self {
```
jonhoo: Same here, see [`with_capacity_and_hasher`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_capacity_and_hasher).

src/lib.rs
```diff
@@ -325,14 +325,44 @@ where
     }
 }
 
+impl<K, V, S: BuildHasher> FlurryHashMap<K, V, S> {
+    /// Creates a new, empty map with the default initial table size (16).
+    /// It uses the given hasher, rather than the default `RandomState`.
+    pub fn new_with_hasher(hasher: S) -> Self {
+        Self {
```
jonhoo: We probably want `fn new` to call this method rather than have both of them separately construct a `Self`.

jonhoo:
Great catch, thanks!

src/lib.rs
```diff
@@ -299,24 +299,50 @@ where
...
+
+    /// Creates a new, empty map with an initial table size accommodating the specified number of
+    /// elements without the need to dynamically resize.
+    pub fn with_capacity(n: usize) -> Self {
+        Self::with_capacity_and_hasher(RandomState::new(), n)
+    }
+}
+
+impl<K, V, S: BuildHasher> FlurryHashMap<K, V, S> {
+    /// Creates an empty HashMap which will use the given hash builder to hash keys.
```
```suggestion
    /// Creates an empty map which will use `hash_builder` to hash keys.
```

src/lib.rs
```diff
@@ -299,24 +299,50 @@ where
...
+    ///
+    /// The created map has the default initial capacity.
+    ///
+    /// Warning: hash_builder is normally randomly generated, and is designed to
```
```suggestion
    /// Warning: `hash_builder` is normally randomly generated, and is designed to
```

src/lib.rs
```diff
@@ -299,24 +299,50 @@ where
...
             size_ctl: AtomicIsize::new(0),
-            build_hasher: RandomState::new(),
+            build_hasher: hash_builder,
         }
     }
 
-    /// Creates a new, empty map with an initial table size accommodating the specified number of
-    /// elements without the need to dynamically resize.
-    pub fn with_capacity(n: usize) -> Self {
+    /// Creates an empty HashMap with the specified capacity, using hash_builder to hash the keys.
```
```suggestion
    /// Creates an empty map with the specified `capacity`, using `hash_builder` to hash the keys.
```

src/lib.rs
```diff
@@ -299,24 +299,50 @@ where
...
+    ///
+    /// The hash map will be able to hold at least capacity elements without reallocating. If
+    /// capacity is 0, the hash map will not allocate.
```
jonhoo: I'm not sure we actually give the _guarantee_ that it will not have to reallocate. If, for example, all the keys has to the same bin, we will resize sooner. It's certainly a degenerate case, but we don't want to promise something we cannot guarantee. How about:

```rust
// The map will be sized to accommodate `capacity` elements with a low chance of reallocating (assuming uniformly distributed hashes). If `capacity` is 0, the call will not allocate, and is equivalent to [`FlurryHashMap::new`].
```

src/lib.rs
```diff
@@ -299,24 +299,50 @@ where
...
+    ///
+    /// Warning: hash_builder is normally randomly generated, and is designed to allow HashMaps
```
```suggestion
    /// Warning: `hash_builder` is normally randomly generated, and is designed to allow the map
```

src/lib.rs
```diff
@@ -299,24 +299,50 @@ where
...
+    }
+}
+
+impl<K, V, S: BuildHasher> FlurryHashMap<K, V, S> {
+    /// Creates an empty HashMap which will use the given hash builder to hash keys.
+    ///
+    /// The created map has the default initial capacity.
+    ///
+    /// Warning: hash_builder is normally randomly generated, and is designed to
+    /// allow HashMaps to be resistant to attacks that cause many collisions and
```
```suggestion
    /// allow the map to be resistant to attacks that cause many collisions and
```

jonhoo:
Haha, sorry, I _just_ merged #25, so you'll need to resolve conflicts with `master` again :sweat_smile:

PR #25, opened 2020-01-24
Allow borrowed keys for lookups

cuviper:
This matches the standard library's approach to key lookups, allowing
any `Q where K: Borrow<Q>`. This lets you use unowned values to query
the maps, e.g. a map with `K = String` can be accessed with `Q = str`.

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -1,6 +1,7 @@
 use super::Table;
 use crossbeam::epoch::{Atomic, Guard, Shared};
 use parking_lot::Mutex;
+use std::borrow::Borrow;
 use std::sync::atomic::Ordering;
 
 /// Entry in a bin.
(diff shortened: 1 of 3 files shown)


cuviper:
I think this is technically a breaking change. Current uses of `&K` will still work since `Borrow` is reflexive, but the added type parameter on the methods would affect anything that explicitly specifies those parameters already.

src/node.rs
```diff
@@ -60,7 +62,7 @@ where
                         unreachable!();
                     };
 
-                    if n.hash == hash && &n.key == key {
+                    if n.hash == hash && n.key.borrow() == key {
```
jonhoo: We can probably do this borrow outside of the loop?

jonhoo:
Beautiful! It is a breaking change, but since we're at `0.0.1` I don't worry too much about that :)

src/node.rs
cuviper: How so? Isn't `n` changing each time?

jonhoo: Ah, yes, of course, you are right. I had `K` and `Q` inverted in my head.

cuviper:
Thanks!

Just as a rule of thumb going forward -- any time any API takes a `&K` parameter, it can probably use `&Q` instead. Same for `&T` when sets are added.

jonhoo:
Yup! This is why #10 mentioned

> Make sure to look at the bounds that `std::collections::HashMap` uses!

:p

Issue #26, opened 2020-01-24
no_std + alloc


jonhoo:
This crate _should_ be possible to write using only [`core`](https://doc.rust-lang.org/core/) and [`alloc`](https://doc.rust-lang.org/alloc/), and thus be usable in `no_std` environments.

The `hashbrown` implementation should be hugely helpful in figuring out how to do this. In particular, see the magical incantations in [`src/lib.rs`](https://github.com/rust-lang/hashbrown/blob/master/src/lib.rs) and the use of `alloc` and `core` in [`src/raw/mod.rs`](https://github.com/rust-lang/hashbrown/blob/master/src/raw/mod.rs). We may be able to simplify some of those since `hashbrown` has to [pull some tricks](https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/Cargo.toml#L23-L26) so that it can be pulled into `std`. Compare for example with the `no_std` support in [`crossbeam-epoch/src/lib.rs`](https://github.com/crossbeam-rs/crossbeam/blob/c0a999d2f15935a895cbcee16f3b46746d5174b7/crossbeam-epoch/src/lib.rs#L59).

There are a couple of things that we have to figure out how to operate in a `no_std` environment:
 - `crossbeam-epoch` will need `default-features = false` and `features = ["alloc"]`. This will [remove `epoch::pin`](https://github.com/crossbeam-rs/crossbeam/blob/c0a999d2f15935a895cbcee16f3b46746d5174b7/crossbeam-epoch/src/lib.rs#L104) (but not `epoch::Guard` I think), and we'll need to figure out whether the API is still usable (and maybe include some examples).
 - `RandomState` is only available with `std`, so we will want to add a `std` feature (which is on by default) and only expose it when the feature is enabled. We'll also have to figure out how that affects the [`default value for S`](https://github.com/jonhoo/flurry/blob/2a5bd258fa805e89715f7dbd52b63bddf736490f/src/lib.rs#L262).
 - Neither `num_cpus` nor `Once` (which we use for [`NCPU`](https://github.com/jonhoo/flurry/blob/2a5bd258fa805e89715f7dbd52b63bddf736490f/src/lib.rs#L247)) are available on `no_std`. We'll need to figure out how to [determine the stride](https://github.com/jonhoo/flurry/blob/2a5bd258fa805e89715f7dbd52b63bddf736490f/src/lib.rs#L837) for transfers without it.
[...]

PR #27, opened 2020-01-24
Move rand to [dev-dependencies]

cuviper:
It's only used in a few tests, not in the crate proper.


jonhoo:
Good catch! I think there's an argument that `num_cpus` should also be moved behind a feature flag. We'd just have to figure out a sensible value to set [`stride`](https://github.com/jonhoo/flurry/blob/67cf13e87b10a1e0db585de136ba54b6fa28b065/src/map.rs#L664) to. That would also help with #26 where `num_cpus` isn't available.

PR #28, opened 2020-01-24
implement "reserve" and improve "Extend" impl

soruh:
This Draft somewhat naively follows the suggestions made at  #10 to implement `reserve` and improve the `Extend` implementation.
I'll continue to work on this and update this PR when I have made more progress

# I have so far:
- naively ported the `tryPresize` function
- implemented a `reserve` wrapper around it (which, like most of the public API, takes a `Guard` for now)
- improved the `Extend` implementation to use `reserve` and the `hashbrown::HashMap::extend` heuristic on how much to reserve
- written some tests, to confirm everything works. (I had to implement `capacity` to make sure a resize actually happened`
- written more tests
- fixed the resizeStamp function
- refactored the load factor calculation

# What still needs to be done:
- review the code
~- possibly relax some `Ordering`s (everything is `SeqConst` for now)~ (Should probably get it's own PR)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -461,7 +461,7 @@ where
                         let n = unsafe { p.deref() }.as_node().unwrap();
                         if n.hash == h && &n.key == key {
                             // the key already exists in the map!
-                            let current_value = head.value.load(Ordering::SeqCst, guard);
+                            let current_value = n.value.load(Ordering::SeqCst, guard);
 
                             // safety: since the value is present now, and we've held a guard from
                             // the beginning of the search, the value cannot be dropped until the
--- a/src/map.rs
+++ b/src/map.rs
@@ -940,58 +940,107 @@ where
 
     /// Tries to presize table to accommodate the given number of elements.
     fn try_presize<'g>(&self, size: usize, guard: &'g Guard) {
-        let new_capacity = if size >= MAXIMUM_CAPACITY / 2 {
+        let requested_capacity = if size >= MAXIMUM_CAPACITY / 2 {
             MAXIMUM_CAPACITY
         } else {
+            // round the requested_capacity to the next power of to from 1.5 * size + 1
+            // TODO: find out if this is neccessary
             let size = size + (size >> 1) + 1;
 
             std::cmp::min(MAXIMUM_CAPACITY, size.next_power_of_two())
-        };
-
-        let c = new_capacity as isize; // TODO
+        } as isize;
 
         loop {
-            let mut sc = self.size_ctl.load(Ordering::SeqCst);
-            if sc < 0 {
+            let size_ctl = self.size_ctl.load(Ordering::SeqCst);
+            if size_ctl < 0 {
                 break;
             }
 
-            let mut table = self.table.load(Ordering::SeqCst, &guard);
+            let table = self.table.load(Ordering::SeqCst, &guard);
 
-            let n = match table.is_null() {
+            // The current capacity == the number of bins in the current table
+            let current_capactity = match table.is_null() {
                 true => 0,
                 false => unsafe { table.deref() }.len(),
             };
 
-            if n == 0 {
-                let n = sc.max(c) as usize;
+            if current_capactity == 0 {
+                // the table has not yet been initialized, so we can just create it
+                // with as many bins as were requested
 
-                if self.size_ctl.compare_and_swap(sc, -1, Ordering::SeqCst) == sc {
-                    if self.table.load(Ordering::SeqCst, guard) == table {
-                        table = Owned::new(Table::new(n)).into_shared(guard);
-                        let old_table = self.table.swap(table, Ordering::SeqCst, &guard);
+                // since the map is uninitialized, size_ctl describes the initial capacity
+                let initial_capacity = size_ctl;
 
-                        assert!(old_table.is_null());
-                        // TODO
-                        // if !old_table.is_null() {
-                        //      unsafe { guard.defer_destroy(old_table) }
-                        // }
+                // the new capacity is either the requested capacity or the initial capacity (size_ctl)
+                let new_capacity = requested_capacity.max(initial_capacity) as usize;
 
-                        sc = n as isize - (n >> 2) as isize;
-                    }
-
-                    self.size_ctl.store(sc, Ordering::SeqCst);
+                // try to aquire the initialization "lock" to indicate that we are initializing the table.
+                if self
+                    .size_ctl
+                    .compare_and_swap(size_ctl, -1, Ordering::SeqCst)
+                    != size_ctl
+                {
+                    // somebody else is already initializing the table (or has already finished).
+                    continue;
                 }
-            } else if c <= sc || n >= MAXIMUM_CAPACITY {
+
+                // we got the initialization `lock`; Make sure the table is still unitialized
+                // (or is the same table with 0 bins we read earlier, althought that should not be the case)
+                if self.table.load(Ordering::SeqCst, guard) != table {
+                    // NOTE: this could probably be `!self.table.load(...).is_null()`
+                    // if we decide that tables can never have 0 bins.
+
+                    // the table is already initialized; Write the `size_ctl` value it had back to it's
+                    // `size_ctl` field to release the initialization "lock"
+                    self.size_ctl.store(size_ctl, Ordering::SeqCst);
+                    continue;
+                }
+
+                // create a table with `new_capacity` empty bins
+                let new_table = Owned::new(Table::new(new_capacity)).into_shared(guard);
+
+                // store the new table to `self.table`
+                let old_table = self.table.swap(new_table, Ordering::SeqCst, &guard);
+
+                // old_table should be `null`, since we don't ever initialize a table with 0 bins
+                // and this branch only happens if table has not yet been initialized or it's length is 0.
+                assert!(old_table.is_null());
+
+                // TODO: if we allow tables with 0 bins. `defer_destroy` `old_table` if it's not `null`:
+                // if !old_table.is_null() {
+                //     // TODO: safety argument, for why this is okay
+                //     unsafe { guard.defer_destroy(old_table) }
+                // }
+
+                // resize the table once it's load factor reaches 75%
+                let new_load_to_resize_at = new_capacity as isize - (new_capacity >> 2) as isize;
+
+                // store the next load at which the table should resize to it's size_ctl field
+                // and thus release the initialization "lock"
+                self.size_ctl.store(new_load_to_resize_at, Ordering::SeqCst);
+            } else if requested_capacity <= size_ctl || current_capactity >= MAXIMUM_CAPACITY {
+                // Either the `requested_capacity` was smaller than or equal to the load we would resize at (size_ctl)
+                // and we don't need to resize, since our load factor will still be acceptable if we don't
+
+                // Or it was larger than the `MAXIMUM_CAPACITY` of the map and we refuse
+                // to resize to an invalid capacity
                 break;
             } else if table == self.table.load(Ordering::SeqCst, &guard) {
-                let rs: isize = Self::resize_stamp(n) << RESIZE_STAMP_SHIFT;
-                // TODO: `rs` is postive even though `resize_stamp` says:
+                // The table is initialized, try to resize it to the requested capacity
+
+                let rs: isize = Self::resize_stamp(current_capactity) << RESIZE_STAMP_SHIFT;
+                // TODO: see #29: `rs` is postive even though `resize_stamp` says:
                 // "Must be negative when shifted left by RESIZE_STAMP_SHIFT"
                 // and since our size_control field needs to be negative
                 // to indicate a resize this needs to be addressed
 
-                if self.size_ctl.compare_and_swap(sc, rs + 2, Ordering::SeqCst) == sc {
+                if self
+                    .size_ctl
+                    .compare_and_swap(size_ctl, rs + 2, Ordering::SeqCst)
+                    == size_ctl
+                {
+                    // someone else already started to resize the table
+                    // TODO: can we `self.help_transfer`?
                     self.transfer(table, Shared::null(), &guard);
                 }
             }
--- a/src/map.rs
+++ b/src/map.rs
@@ -12,12 +12,18 @@ use std::sync::{
     Once,
 };
 
+macro_rules! isize_bits {
+    () => {
+        std::mem::size_of::<isize>() * 8
+    };
+}
+
 /// The largest possible table capacity.  This value must be
 /// exactly 1<<30 to stay within Java array allocation and indexing
 /// bounds for power of two table sizes, and is further required
 /// because the top two bits of 32bit hash fields are used for
 /// control purposes.
-const MAXIMUM_CAPACITY: usize = 1 << 30;
+const MAXIMUM_CAPACITY: usize = 1 << 30; // TODO: use isize_bits!()
 
 /// The default initial table capacity.  Must be a power of 2
 /// (i.e., at least 1) and at most `MAXIMUM_CAPACITY`.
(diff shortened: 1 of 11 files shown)


jonhoo:
Awesome! I'll take a look through and leave comments.

src/map.rs
```diff
@@ -937,6 +937,74 @@ where
...
+
+        loop {
+            let sc = self.size_ctl.load(Ordering::SeqCst);
+            if sc < 0 {
+                break;
+            }
+
+            let table = self.table.load(Ordering::SeqCst, &guard);
+
+            // TODO: clean this up?
```
jonhoo: I propose using `0` instead of `None`. That way the check below can simply be `if n == 0`. You could possibly simplify it further with:
```rust
let mut n = 0;
if !table.is_null() {
    // safety: table was read under guard, which still lives, so table cannot be dropped this epoch
    n = unsafe { table.deref() }.len();
}
```

src/map.rs
```diff
@@ -937,6 +937,74 @@ where
...
+            let n = if table.is_null() {
+                None
+            } else {
+                Some(unsafe { table.deref() }.len())
+            };
+
+            if n.is_none() || n.unwrap() == 0 {
+                let n = sc.max(c) as usize;
+
+                let mut table = self.table.load(Ordering::SeqCst, guard);
```
jonhoo: I think this read of `table` has to come _after_ the CAS of `size_ctl`. These things can matter a lot in concurrent code!

jonhoo: I think this line should not be here. The comparison inside the `if` below is specifically against the `table` that is read above.

src/map.rs
```diff
@@ -937,6 +937,74 @@ where
...
+
+                if self.size_ctl.compare_and_swap(sc, -1, Ordering::SeqCst) == sc {
+                    if self.table.load(Ordering::SeqCst, guard) == table {
+                        let new_table = Owned::new(Table::new(n));
+                        table = new_table.into_shared(guard);
```
jonhoo: I don't think you need an intermediate binding here -- just call `into_shared` directly in `store`.

src/map.rs
soruh: That's a great idea, what do you think of something like this:
```rust
let n = match table.is_null() {
    true => 0,
    false => unsafe { table.deref() }.len(),
};
```

soruh: That way we wouldn't need the `mut n`, even though we don't really mutate it

src/map.rs
```diff
@@ -937,6 +937,74 @@ where
...
+                        self.table.store(table, Ordering::SeqCst);
```
jonhoo: _Technically_ this should use `swap` instead, and the old table should be marked as garbage. See [this code in `transfer`](https://github.com/jonhoo/flurry/blob/36ddd9285f8c91b4a82591582f19c61505f84638/src/map.rs#L721). Now, realistically we never (I don't think) allocate a table with 0 bins, so by arriving here (thanks to the `if` above) it is most likely that `self.table` _was_ `null`. We should _either_ `assert!` that the old table value was `null` here, or we should properly mark a non-`null` result from `.swap` as garbage.

src/map.rs
```diff
@@ -937,6 +937,74 @@ where
...
+                        self.size_ctl
```
jonhoo: This should change `sc`, not `self.size_ctl` directly. `self.size_ctl` is modified just below the `if`.

src/map.rs
soruh: Your right, I'll remove it

src/map.rs
soruh: Nice, I automatically assumed the borrow checker wouldn't like that, but since we're that low-level I guess it doesn't really care

src/map.rs
```diff
@@ -937,6 +937,74 @@ where
...
+            }
+        }
+    }
+
+    #[inline]
+    /// Tries to reserve capacity for at least additional more elements.
+    /// The collection may reserve more space to avoid frequent reallocations.
+    pub fn reserve(&self, additional: usize) {
+        let guard = epoch::pin();
+        self.try_presize(additional, &guard);
```
jonhoo: I _think_ `try_presize` takes the _absolute_ table size as an argument, whereas `reserve` takes `additional`, so this won't work out quite right.

src/map.rs
```diff
@@ -1126,6 +1194,20 @@ where
         self.count.load(Ordering::Relaxed)
     }
 
+    #[inline]
+    /// Returns the capacity of the map.
+    pub fn capacity<'g>(&self, guard: &'g Guard) -> usize {
```
jonhoo: I'm actually not sure we want to expose this method publicly. There (probably) aren't any good reasons why callers should care how many bins there were, especially given concurrency, so I think I'd just make this private. And maybe even `#[cfg(test)]`.

src/map.rs
jonhoo: Beautiful!

src/map.rs
soruh: You're totally correct, I'll fix that now

src/map.rs
soruh: I just added it for now to be able to write the tests and to get your opinion on this. I think `#[cfg(test)]` would be the most reasonable

src/map.rs
soruh: I think I'll just`assert` for now and see about marking the result as garbage later

src/map.rs
soruh: Oh right, I totally forgot about that.
But converting from additional to absolute shouldn't be that hard
```rust
let absolute = self.len() + additional;
```
would do it, right?

jonhoo: Yes, that should be right. It's racy, but then again, that's essentially what the user requested!

src/map.rs
soruh: it's not actually that easy as just slapping a `#[cfg(test)]` on there, since the tests in `tests` don't trigger that. (I assume, at least cargo check complained about not being able to find `capacity`).
I'll see if I can sort this out tomorrow

src/map.rs
```diff
@@ -951,40 +951,41 @@ where
...
 
         loop {
-            let sc = self.size_ctl.load(Ordering::SeqCst);
+            let mut sc = self.size_ctl.load(Ordering::SeqCst);
             if sc < 0 {
                 break;
             }
 
-            let table = self.table.load(Ordering::SeqCst, &guard);
+            let mut table = self.table.load(Ordering::SeqCst, &guard);
```
jonhoo: I don't think this has to be mutable? You never use the value later, so it doesn't need to be re-assigned.

src/map.rs
```diff
@@ -951,40 +951,41 @@ where
...
-                let mut table = self.table.load(Ordering::SeqCst, guard);
-
                 if self.size_ctl.compare_and_swap(sc, -1, Ordering::SeqCst) == sc {
                     if self.table.load(Ordering::SeqCst, guard) == table {
-                        let new_table = Owned::new(Table::new(n));
-                        table = new_table.into_shared(guard);
-                        self.table.store(table, Ordering::SeqCst);
-                        self.size_ctl
-                            .store(n as isize - (n >> 2) as isize, Ordering::SeqCst);
+                        table = Owned::new(Table::new(n)).into_shared(guard);
```
```suggestion
                        let new_table = Owned::new(Table::new(n)).into_shared(guard);
```

src/map.rs
```diff
@@ -951,40 +951,41 @@ where
...
+                        let old_table = self.table.swap(table, Ordering::SeqCst, &guard);
```
```suggestion
                        let old_table = self.table.swap(new_table, Ordering::SeqCst, &guard);
```

src/map.rs
```diff
@@ -951,40 +951,41 @@ where
...
+
+                        assert!(old_table.is_null());
+                        // TODO
+                        // if !old_table.is_null() {
+                        //      unsafe { guard.defer_destroy(old_table) }
+                        // }
```
jonhoo: Maybe add brief note about why we think it's probably always `null` here?

src/map.rs
jonhoo: Ah, yes, you're right.. That's awkward. I guess leave it as `pub` then :man_shrugging:

src/map.rs
soruh: Right, I did it this way for the moment, since that is what the java version did, but you're totally correct in that we can just bind to a new variable. I'll change this in a bit.

jonhoo: Ah, but keep in mind that in the Java version, `table` refers to the equivalent of `self.table`, it is _not_ a local variable!

soruh: You're right, I still get confused with the automatic `this.` sometimes, haven't done too much java.
This shouldn't be a problem, since our `table` is the java implementations `tab` which is a local and `new_table` is `nt` which is also a local.

What is your opinion on variable names btw.? I would like to give the variables more expressive names, but would also be okay with staying as close to the java version as possible...

soruh:
I am a bit confused about why the java version multiplies the requested capacity
by 1.5 and adds 1 before rounding it to  the next power of two:
```rust
tableSizeFor(size + (size >>> 1) + 1)
```
and why it set's the size control to 3/4 of the current capacity if the table was not yet initialized: 
```rust 
sc = n - (n >>> 2);
```
It obviously is the right thing to do since the java version does it this way, but it seems really weird to me...

src/map.rs
jonhoo: Hehe, yeah, it can definitely get confusing. You're right that the code is still correct, but when writing unsafe code like this, I like to not mutate pointers or keep them around any more than I have to. In this case, there is no reason for the (local) variable `table` to be `mut`, and therefore it shouldn't be. It's just to prevent ourselves from doing something stupid down the line :sweat_smile: 

As for variable names, I'm totally fine with _either_ using the exact same variable names as the Java code, or coming up for _good_ descriptive names. The only thing I wouldn't like is giving a variable a different but still relatively obscure/generic name :)

jonhoo:
> I am a bit confused about why the java version multiplies the requested capacity
by 1.5 and adds 1 before rounding it to the next power of two

I'm not sure. It may have to do with the way that their `tableSizeFor` function is implemented, where you get rounded to the _nearest_ power of 2 rather than the _next_ power of two?

> Why it set's the size control to 3/4 of the current capacity if the table was not yet initialized

`size_ctl` is set to 3/4 of the _new_ capacity _if_ we end up allocating. Think of this as "when the next resize should happen". ConcurrentHashMap specifically targets a 0.75 load factor, and resizes if it goes above that so that no single bin gets too crowded.

soruh:
Ah, I was confused about the exact meaning of size_ctl, guess I'll look at the rest of my code, to make sure I didn't make any wrong comments because of that.

soruh:
since if `size_ctl` is 0 it means the inital capacity should be `DEFAULT_CAPACITY` shouldn't this line:
```rust
n = (sc > c) ? sc : c;
```
take that into account?

soruh:
I'm thinking of something like this:
```rust
let initial_capacity = if size_ctl == 0 {
    DEFAULT_CAPACITY as isize
} else {
    size_ctl
};

// the new capacity is either the requested capacity or the initial capacity, if it is larger
// since all maps should be at least as big as their initial capacity specifies
let new_capacity = requested_capacity.max(initial_capacity) as usize;
```

jonhoo:
Well, the initial capacity for a table is determined here:
https://github.com/jonhoo/flurry/blob/051ca79fe2458af0e0e75ab93f8cd113078cd2bd/src/map.rs#L313-L317

`DEFAULT_CAPACITY` shouldn't be used in `reserve`, because there the user already provides us with a target size.

Or maybe I'm missing something about your question?

soruh:
We'll if I used a map and initialized it with an initial capacity I would assume that that it has at least that capacity...

jonhoo:
Ohhh, I see what you're getting at:
```rust
let map = flurry::HashMap::with_capacity(1024);
map.reserve(512);
```
The call to `with_capacity` just sets `size_ctl`, but it does not allocate anything. So when we get to `reserve`, there will be no bins, but `size_ctl != 0`. I think your approach of taking the max is probably the right one. It also makes me wonder if we should call `.reserve` from `with_capacity` automatically so that we _actually_ allocate when the user uses `with_capacity`.

jonhoo:
b83cbe279e711d11c59a05c29afdb47d52337e8f looks great

soruh:
I think calling `try_presize` from `with_capacity` would be the way to go, but I realized, that the `DEFAULT_CAPACITY` isn't really needed.
`size_ctl` is only 0 if the user _didn't_ specify a capacity and in that case, since the user doesn't care what the initial capacity was, we can just use the requested capacity...

soruh:
The question is how we handle the guard for `try_presize` in `with_capacity`. I think we can just use an `unprotected`, since we are creating the map and no one else can access it, but I'm not completely certain.

soruh:
Is there a reason you were dividing by the load factor:
```rust
let size = (1.0 + (n as f64) / LOAD_FACTOR) as usize;
```
instead of multiplying by it:
```rust
let size = (1.0 + (n as f64) * LOAD_FACTOR) as usize;
```

or was that just a typo?

src/map.rs
soruh: I just looked at this a bit more and if we make tests that use `capacity` unit tests (so move them to the same module) It'll work, so I'll do that.

soruh:
I initially wanted to make a new branch for this but impulsively committed this here, should I revert the last commit and make a new branch, or can I just leave it?

jonhoo:
Well, `DEFAULT_CAPACITY` is the value you should allocate if the user does _not_ specify a capacity. That is, if neither `with_capacity` nor `reserve` is called.

It should be fine to use `unprotected` as the guard in `with_capacity`.

I divide because that's what the Java code [does](https://github.com/jonhoo/flurry/blob/051ca79fe2458af0e0e75ab93f8cd113078cd2bd/jsr166/src/ConcurrentHashMap.java#L869).

I'm fine with you folding the fix for the resize stamp in here :)

jonhoo:
This is looking really good. Are there any other changes you want to make before merging?

soruh:
> Well, `DEFAULT_CAPACITY` is the value you should allocate if the user does _not_ specify a capacity. That is, if neither `with_capacity` nor `reserve` is called.

But we don't need it in `try_presize`, since we only call it when the user _has_ called `with_capacity` or `reserve`

> I divide because that's what the Java code [does](https://github.com/jonhoo/flurry/blob/051ca79fe2458af0e0e75ab93f8cd113078cd2bd/jsr166/src/ConcurrentHashMap.java#L869).

That's really weird, especially since it adds a global constant just for that...
Since we call `try_presize` in `with_capacity` now, we don't have that line anymore, since `try_presize` manages the load factor; I don't think this will be a problem though.

> I'm fine with you folding the fix for the resize stamp in here :)

Great, thank you

soruh:
> This is looking really good. Are there any other changes you want to make before merging?

Not really, I initially wanted to look into maybe relaxing some `Ordering`s, but I think it might be better to do that in a new PR, since we also need to do that in other parts of the code...

jonhoo:
> But we don't need it in `try_presize`, since we only call it when the user has called `with_capacity` or `reserve`

Right, hence my comment above:
> `DEFAULT_CAPACITY` shouldn't be used in `reserve`, because there the user already provides us with a target size.

---

> I initially wanted to look into maybe relaxing some `Ordering`s, but I think it might be better to do that in a new PR, since we also need to do that in other parts of the code...

Yes, I think that belongs in a new PR!

Okay, let's merge this :tada:

soruh:
> > But we don't need it in `try_presize`, since we only call it when the user has called `with_capacity` or `reserve`
> 
> Right, hence my comment above:
> 
> > `DEFAULT_CAPACITY` shouldn't be used in `reserve`, because there the user already provides us with a target size.
>

Whoops, guess I missed that, sorry.

> Okay, let's merge this 🎉

Great, Thank you!

jonhoo:
No, thank you!

Issue #12, opened 2020-01-23
Add reservation-based methods


jonhoo:
Add [`computeIfAbsent`], [`computeIfPresent`], [`compute`]

I have a suspicion that `ReservationNode` could also be used to implement an [`Entry`-API] like the one on `std::collections::HashMap`, so that may be worth looking into.

  [`computeIfAbsent`]: https://github.com/jonhoo/flurry/blob/d3dae0465b37b7f12c4f0d58a16f36fb1d8c1596/jsr166/src/ConcurrentHashMap.java#L1662
  [`computeIfPresent`]: https://github.com/jonhoo/flurry/blob/d3dae0465b37b7f12c4f0d58a16f36fb1d8c1596/jsr166/src/ConcurrentHashMap.java#L1774
  [`compute`]: https://github.com/jonhoo/flurry/blob/62dcd58b5ee47ed8ee51086bb9dc84b2c499fb6c/jsr166/src/ConcurrentHashMap.java#L1868
  [`Entry`-API]: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry

voidc:
I started working on this in #32. For now, only compute_if_present is implemented.

jhinch:
I would like to propose the following for the entry API:

```rust
impl<K, V, S> HashMap<K, V, S> {
    pub fn entry<'g>(&'g self, key: K, guard: &'g Guard) -> Entry<'g, K, V, S> {
        todo!()
    }
}

pub enum Entry<'g, K, V, S> {
    Vacant(VacantEntry<'g, K, V, S>),
    Occupied(OccupiedEntry<'g, K, V, S>),
}

pub struct VacantEntry<'g, K, V, S> {
    _todo: PhantomData<('g, K, V, S)>,
}

pub struct OccupiedEntry<'g, K, V, S> {
    _todo: PhantomData<('g, K, V, S)>,
}

impl<'g, K, V, S> Entry<'g, K, V, S> {
    pub fn or_insert(self, default: V) -> &'g V {
        todo!()
    }

    pub fn or_insert_with<F: FnOnce() -> V>(self, default: F) -> &'g V {
        todo!()
    }

    pub fn key(&'g self) -> &'g K {
        todo!()
    }

    pub fn and_modify<F: FnOnce(&'g V) -> V>(self, f: F) -> Self {
        todo!()
    }
}

impl<'g, K, V: Default, S> Entry<'g, K, V, S> {
[...]

Issue #9, opened 2020-01-23
Finish the safety argument for `BinEntry::Moved`


jonhoo:
The safety argument for why dereferencing the pointer in `BinEntry::Moved` has not been completed. It currently stands with a `FIXME` in case three here
https://github.com/jonhoo/flurry/blob/d3dae0465b37b7f12c4f0d58a16f36fb1d8c1596/src/lib.rs#L474-L492
and case two here
https://github.com/jonhoo/flurry/blob/6d56c58026f6edd0d9828ce48b3cbb6aecd38973/src/lib.rs#L1060-L1072

domenicquirl:
I'm thinking as follows: If we perform an operation that sees a `Moved(t)`, we must have gotten there by _loading_ `self.table` with a guard `g`. This guard `g` will pin an epoch _before_ the resize is finished with respect to the loaded table, otherwise we would have loaded the next table. This argument also exists here:
https://github.com/jonhoo/flurry/blob/051ca79fe2458af0e0e75ab93f8cd113078cd2bd/src/map.rs#L719-L749

- During the resize, `t` points to `next_table` and is thus valid.
- After the resize, `t == self.table`, thus `t` is valid.
- This holds until a subsequent resize ends, at which point `self.table != t` and `t` is `defer_destroy`ed (see the code above). At this point, `t` is not referenced by the map anymore. However, the `defer_destroy` happens while `g` is still pinning the epoch. Thus, `t` remains valid for at least the lifetime of `g`.
- After releasing `g`, performing operations on the map with a new guard `g'` cannot access the table anymore (also see above).

Since this traces the entire lifetime of both the guard and the table and in particular, the above code is the only place that `defer_destroy`s tables, I think the argument is sound. What do you think?

jonhoo:
Yes, that seems correct to me as well! Could you submit a PR with that argument included? I think maybe we want to make that argument as a comment on the `Moved` variant of `BinEntry` (rather than in everywhere that uses it), and then reference that argument from the two or three places that rely on it.

domenicquirl:
Nice! I just created #35 for this. I was also already thinking about reducing redundancy with these comments, but wasn't sure where to put them yet. The definition of `Moved` seems like a good place!

There is also some other PR i saw which had copied the argument, I'll go find it and tell them about the PR so that we don't introduce more copies of the FIXME again 😉

PR #32, opened 2020-01-25
Implement compute_if_present

voidc:
This pull request contains the compute_if_present method, which is the simplest one among the reservation-based methods mentioned in #12 since it does not require ReservationNodes.

I took most of the code from put and remove.
For now, I only added basic tests but I also want to port some of the Java tests.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -535,14 +535,16 @@ where
     /// Maps `key` to `value` in this table.
     ///
     /// The value can be retrieved by calling [`get`] with a key that is equal to the original key.
-    pub fn compute_if_present<'g, F>(
+    pub fn compute_if_present<'g, Q, F>(
         &'g self,
-        key: K,
+        key: &Q,
         remapping_function: F,
         guard: &'g Guard,
     ) -> Option<&'g V>
     where
-        F: Fn(&K, &V) -> Option<V>,
+        K: Borrow<Q>,
+        Q: ?Sized + Hash + Eq,
+        F: FnOnce(&K, &V) -> Option<V>,
     {
         let h = self.hash(&key);
 
@@ -639,7 +641,7 @@ where
                         let n = unsafe { p.deref() }.as_node().unwrap();
                         // TODO: This Ordering can probably be relaxed due to the Mutex
                         let next = n.next.load(Ordering::SeqCst, guard);
-                        if n.hash == h && &n.key == &key {
+                        if n.hash == h && n.key.borrow() == key {
                             // the key already exists in the map!
                             let current_value = head.value.load(Ordering::SeqCst, guard);
 
@@ -648,7 +650,7 @@ where
                             // next epoch, which won't arrive until after we drop our guard.
                             let current_value = unsafe { current_value.deref() };
 
-                            let new_value = remapping_function(&key, current_value);
+                            let new_value = remapping_function(&n.key, current_value);
                             //https://youtu.be/yQFWmGaFBjk?t=10800
                             if let Some(value) = new_value {
                                 // safety: we own value and have never shared it
--- a/src/map.rs
+++ b/src/map.rs
@@ -543,6 +543,9 @@ where
     /// attempted update operations on this map by other threads may be
     /// blocked while computation is in progress, so the computation
     /// should be short and simple.
+    /// 
+    /// Returns the new value associated with the specified `key`, or `None`
+    /// if no value for the specified `key` is present.
     pub fn compute_if_present<'g, Q, F>(
         &'g self,
         key: &Q,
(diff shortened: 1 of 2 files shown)


src/map.rs
```diff
@@ -532,6 +532,198 @@ where
...
+                            let current_value = head.value.load(Ordering::SeqCst, guard);
+
+                            // safety: since the value is present now, and we've held a guard from
+                            // the beginning of the search, the value cannot be dropped until the
+                            // next epoch, which won't arrive until after we drop our guard.
+                            let current_value = unsafe { current_value.deref() };
+
+                            let new_value = remapping_function(&n.key, current_value);
+                            if let Some(value) = new_value {
+                                // safety: we own value and have never shared it
```
jonhoo: You don't actually need this `// safety`, since there's nothing `unsafe` about the next statement. There is in `put` where we need to promise that the value we take out of the `Node` is actually one we own.

src/map.rs
```diff
@@ -532,6 +532,198 @@ where
...
+                                        .next
+                                        .store(next, Ordering::SeqCst);
+                                } else {
+                                    // or by setting the next node as the first BinEntry if there is no previous entry
+                                    t.store_bin(bini, next);
+                                }
+
+                                // in either case, mark the BinEntry as garbage, since it was just removed
+                                // safety: see remove
+                                unsafe { guard.defer_destroy(p) };
```
jonhoo: In this case you _also_ need to `defer_destroy` the `.value` _inside_ `p`. Notice how `remove` does that [here](https://github.com/jonhoo/flurry/blob/a5e7cea119e6fd29bc1ab5748c35c0987f81df90/src/map.rs#L1079).

src/map.rs
```diff
@@ -532,6 +532,198 @@ where
...
+                            // we're at the end of the bin
+                            break None;
+                        }
+                        p = next;
+
+                        bin_count += 1;
+                    };
+                    drop(head_lock);
+
+                    // TODO: TREEIFY_THRESHOLD
```
jonhoo: It doesn't seem like the Java code considers treeifying in `compute_if_present`, so this comment can go away.

src/map.rs
```diff
@@ -532,6 +532,198 @@ where
...
+
+                    if delta != 0 {
```
jonhoo: I would probably replace `delta` with `removed_node: bool`, and then call `add_count(-1, ...)`.

jonhoo:
This looks quite good! I left a couple of inline notes.

It makes me want to try to figure out how we might be able to unify the code for `put`, `compute_if_present`, `remove`, and maybe even some others. They all have roughly the same "outer" code.. But that's for another time.

jonhoo:
@voidc Hmm, why do you want the semantics to be that it returns the new value rather than the old? Or, I guess, it could return both?

voidc:
Hi! Thanks for the comments!
I changed the function to return the computed, new value so it matches the Java version.
But your are right, it could return both.

jonhoo:
Ohh, I see, I missed that! Yes, I think returning references to both old and new would be good.

src/map.rs
```diff
@@ -532,6 +532,205 @@ where
...
+                        // safety: we read the bin while pinning the epoch. a bin will never be
+                        // dropped until the next epoch after it is removed. since it wasn't
+                        // removed, and the epoch was pinned, that cannot be until after we drop
+                        // our guard.
+                        let n = unsafe { p.deref() }.as_node().unwrap();
+                        // TODO: This Ordering can probably be relaxed due to the Mutex
+                        let next = n.next.load(Ordering::SeqCst, guard);
+                        if n.hash == h && n.key.borrow() == key {
+                            // the key already exists in the map!
+                            let current_value = head.value.load(Ordering::SeqCst, guard);
```
voidc: Why is `current_value` is loaded from `head.value` instead of `n.value`?

jonhoo: That is an excellent question! I believe that is a bug. I'll push a fix and a unit test for that shortly!

jonhoo: Fixed in https://github.com/jonhoo/flurry/commit/051ca79fe2458af0e0e75ab93f8cd113078cd2bd

src/map.rs
voidc: Is this resolved by https://github.com/jonhoo/flurry/commit/729c82f571141f65344f4a2e89d5a64efc7c419e?

jonhoo: Yup!

jonhoo: Yes, although it'd be good to also add a short note about why we also need to free the value.

domenicquirl:
@voidc Hi, I just submitted #35 to deal with #9 and noticed you are also matching against `BinEntry::Moved` and have copied over the corresponding safety comment. Would you mind to change the comment to reference the new safety statement (it's now _on_ the definition of `Moved`) so we don't spread the `FIXME` further? 😊

jonhoo:
@voidc Unless there's more you want to do with this, I think I'm happy to merge once you bring it up to date with `master`!

voidc:
Thanks for the approval! Feel free to merge.
Points of (future) consideration:

- Also return old value as (although this would mean being inconsistent with the other compute_* methods as they don't necessarily have an old value)
- Possibly unify code with `replace_node` and `put`.

PR #35, opened 2020-01-26
Finish safety argument for BinEntry::Moved

domenicquirl:
This tackles #9 and moves the argument to the definition of BinEntry::Moved, so as to reduce redundancy.


jonhoo:
Excellent, thank you!

PR #36, opened 2020-01-26
retain method

danielSanchezQ:
Implemented `retain` for HashMap
Implemented `retain` tests for empty map, retain all elements, retain none elements and retain arbitrary elements

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -945,26 +945,27 @@ where
     /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
     /// form must match those for the key type.
     pub fn remove<'g, Q>(&'g self, key: &Q, guard: &'g Guard) -> Option<&'g V>
-        where
-            K: Borrow<Q>,
-            Q: ?Sized + Hash + Eq,
+    where
+        K: Borrow<Q>,
+        Q: ?Sized + Hash + Eq,
     {
         self.replace_node(key, None, None, guard)
     }
 
     /// Replaces node value with v, conditional upon match of cv.
-    /// If resulting value does not exists it removes the key (and its corresponding value) from this map.
+    /// If resulting value does not exist it removes the key (and its corresponding value) from this map.
     /// This method does nothing if the key is not in the map.
     /// Returns the previous value associated with the given key.
     ///
     /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
     /// form must match those for the key type.
-    pub fn replace_node<'g, Q>(
+    fn replace_node<'g, Q>(
         &'g self,
         key: &Q,
-        new_value : Option<V>,
-        old_value : Option<Shared<'g, V>>,
-        guard: &'g Guard) -> Option<&'g V>
+        new_value: Option<V>,
+        observed_value: Option<Shared<'g, V>>,
+        guard: &'g Guard,
+    ) -> Option<&'g V>
     where
         K: Borrow<Q>,
         Q: ?Sized + Hash + Eq,
@@ -1045,15 +1046,17 @@ where
                         let next = n.next.load(Ordering::SeqCst, guard);
                         if n.hash == h && n.key.borrow() == key {
                             let ev = n.value.load(Ordering::SeqCst, guard);
-                            // found the node but we have a new value to replace the old one
-                            if let Some(nv) = new_value {
-                                n.value.store(Owned::new(nv), Ordering::SeqCst);
-                                break;
+
                             // just remove the node if the value is the one we expected at method call
-                            } else if old_value.is_none() || old_value.unwrap() == ev {
+                            if observed_value.map(|ov| ov == ev).unwrap_or(true) {
+                                // found the node but we have a new value to replace the old one
+                                if let Some(nv) = new_value {
+                                    n.value.store(Owned::new(nv), Ordering::SeqCst);
+                                    break;
+                                }
                                 old_val = Some(ev);
                                 // remove the BinEntry containing the removed key value pair from the bucket
-                               if !pred.is_null() {
+                                if !pred.is_null() {
                                     // either by changing the pointer of the previous BinEntry, if present
                                     // safety: as above
                                     unsafe { pred.deref() }
@@ -1120,16 +1123,37 @@ where
     /// Retains only the elements specified by the predicate.
     ///
     /// In other words, remove all pairs (k, v) such that f(&k,&v) returns false.
+    ///
+    /// This method relies on [`HashMap::replace_node`] and do not force entries deletion if the
+    /// entry value is updated concurrently.
     pub fn retain<F>(&mut self, mut f: F)
-        where
-            F: FnMut(&K, &V) -> bool
+    where
+        F: FnMut(&K, &V) -> bool,
     {
         let guard = epoch::pin();
         // removed selected keys
         for (k, v) in self.iter(&guard) {
             if !f(k, v) {
-                // TODO: right now this removing the element by key without checking the value. The value should be used somehow in old_value
-                self.replace_node(k, None, None, &guard);
+                let old_value: Shared<'_, V> = Shared::from(v as *const V);
+                self.replace_node(k, None, Some(old_value), &guard);
+            }
+        }
+    }
+
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs (k, v) such that f(&k,&v) returns false.
+    ///
+    /// This method force entry deletion even if if it is updated concurrently.
+    pub fn retain_force<F>(&mut self, mut f: F)
+        where
+            F: FnMut(&K, &V) -> bool,
+    {
+        let guard = epoch::pin();
+        // removed selected keys
+        for (k, v) in self.iter(&guard) {
+            if !f(k, v) {
+                self.remove(k, &guard);
             }
         }
     }
@@ -1437,5 +1461,54 @@ fn num_cpus() -> usize {
 /// drop(guard);
 /// drop(r);
 /// ```
+
+#[test]
+fn replace_empty() {
+    let map = HashMap::<usize, usize>::new();
+
+    {
+        let guard = epoch::pin();
+        let old = map.replace_node(&42, None, None, &guard);
+        assert!(old.is_none());
+    }
+}
+
+#[test]
+fn replace_existing() {
+    let map = HashMap::<usize, usize>::new();
+    {
+        let guard = epoch::pin();
+        map.insert(42, 42, &guard);
+        let old = map.replace_node(&42, Some(10), None, &guard);
+        assert!(old.is_none());
+        assert_eq!(*map.get(&42, &guard).unwrap(), 10);
+    }
+}
+
+#[test]
+fn replace_existing_observed_value_matching() {
+    let map = HashMap::<usize, usize>::new();
+    {
+        let guard = epoch::pin();
+        map.insert(42, 42, &guard);
+        let observed_value = Shared::from(map.get(&42, &guard).unwrap() as *const _);
+        let old = map.replace_node(&42, Some(10), Some(observed_value), &guard);
+        assert!(old.is_none());
+        assert_eq!(*map.get(&42, &guard).unwrap(), 10);
+    }
+}
+
+#[test]
+fn replace_existing_observed_value_non_matching() {
+    let map = HashMap::<usize, usize>::new();
+    {
+        let guard = epoch::pin();
+        map.insert(42, 42, &guard);
+        let old = map.replace_node(&42, Some(10), Some(Shared::null()), &guard);
+        assert!(old.is_none());
+        assert_eq!(*map.get(&42, &guard).unwrap(), 42);
+    }
+}
+
 #[allow(dead_code)]
 struct CompileFailTests;
--- a/src/map.rs
+++ b/src/map.rs
@@ -945,6 +945,26 @@ where
     /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
     /// form must match those for the key type.
     pub fn remove<'g, Q>(&'g self, key: &Q, guard: &'g Guard) -> Option<&'g V>
+        where
+            K: Borrow<Q>,
+            Q: ?Sized + Hash + Eq,
+    {
+        self.replace_node(key, None, None, guard)
+    }
+
+    /// Replaces node value with v, conditional upon match of cv.
+    /// If resulting value does not exists it removes the key (and its corresponding value) from this map.
+    /// This method does nothing if the key is not in the map.
+    /// Returns the previous value associated with the given key.
+    ///
+    /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
+    /// form must match those for the key type.
+    pub fn replace_node<'g, Q>(
+        &'g self,
+        key: &Q,
+        new_value : Option<V>,
+        old_value : Option<Shared<'g, V>>,
+        guard: &'g Guard) -> Option<&'g V>
     where
         K: Borrow<Q>,
         Q: ?Sized + Hash + Eq,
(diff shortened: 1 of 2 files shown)


src/map.rs
```diff
@@ -1091,6 +1091,30 @@ where
         None
     }
 
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs (k, v) such that f(&k,&v) returns false.
+    pub fn retain<F>(&mut self, mut f: F)
+        where
+            F: FnMut(&K, &V) -> bool
```
danielSanchezQ: [`std::HashMap::retain`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain) uses `FnMut(&K, &mut V) -> bool` for filtering the map. But in our case we do not have a `get_mut` or similar method, so I changed the signature. I'm not sure if this is ok.

danielSanchezQ:
I used an approach similar to the clone method. I just take the keys that do not pass the filtering function and remove each of them one by one. 
How better/worst would it be to rebuild the map using the `from_iterator` method? (it was my other option, but I went to most basic one). 
I think it depends of the worst case and the amount of re-allocations of the new map.

src/map.rs
jonhoo: No, `(&K, &V)` seems fine since we are working with a concurrent data structure after all!

jonhoo:
I think we probably do not want to construct a new map (with `from_iterator`). That would likely be slower, unless the reduction method was particularly selective, which we don't know in advance. If it was, then the user could easily enough implement that themselves, so the approach you've chosen seems good.

I would probably avoid the intermediate `Vec` though. Since we have concurrency, it should be fine to just do
```rust
for (k, v) in self.iter() {
    if !f(k, v) {
        self.remove(k);
    }
}
```
It looks like hashbrown does something [special about `Drop`](https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L952-L955) in their implementation, but I don't think that affects us since we do not immediately call `drop` regardless. The Java versions ([`retainAll`](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L4538) and [`removeValueIf`](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1621)) also appear to take the same strategy, so I think that's reason enough to stick with it.

One thing to note is that the Java code does this:
https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1630

That is, it uses the `cv` argument of [`replaceNode`](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1081) to _only_ remove the item if the value hasn't changed. We should probably do the same (and document that behavior on `retain`). It _should_  be a matter of:
 1. renaming `remove` to `replace_node`
 2. adding the `v` and `cv` arguments with type `Option<V>` and probably better name
 3. adding a new `remove` that just calls `replace_node(k, None, None, guard)`
 4. adding [this condition](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1103) so that the right thing happens if `cv.is_some()`
 5. for completeness, adding [this condition](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1106-L1107) and making sure we don't unlink the node if `v.is_some()`

danielSanchezQ:
@jonhoo I'm having issues implementing it that way:

```
let n = unsafe { e.deref() }.as_node().unwrap();
let next = n.next.load(Ordering::SeqCst, guard);
if n.hash == h && n.key.borrow() == key {
    let ev = n.value.load(Ordering::SeqCst, guard);
    if new_value.is_none() || old_value.unwrap() == unsafe {ev.deref()} {
        old_val = Some(ev);
        // remove the BinEntry containing the removed key value pair from the bucket
        if new_value.is_some() {
            n.value.store(Owned::new(new_value.unwrap()), Ordering::SeqCst);
        } else if !pred.is_null() {
            // either by changing the pointer of the previous BinEntry, if present
            // safety: as above
            unsafe { pred.deref() }
                .as_node()
                .unwrap()
                .next
                .store(next, Ordering::SeqCst);
        } else {
            // or by setting the next node as the first BinEntry if there is no previous entry
            t.store_bin(i, next);
        }

        // in either case, mark the BinEntry as garbage, since it was just removed
        // safety: as for val below / in put
        unsafe { guard.defer_destroy(e) };

        // since the key was found and only one node exists per key, we can break here
        break;
    }
}
```

Mostly here `if new_value.is_none() || old_value.unwrap() == unsafe {ev.deref()}`, related to [this condition](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1103)  (maybe the other condition too).  I'm not sure how to check if the old_value and the value in the node are the same (I reach the point on where it asks for `V` to implement `PartialEq`, which makes sense). But I feel I may be missing something important

jonhoo:
Ah, the trick there is to not take `&V` as `old_value`, but take `Shared<'g, V>` instead. You want to compare the _pointer_ to the old value to the _pointer_ to the current value (i.e., no `deref`). That would also mean you do not need `V: PartialEq`.

Also, keep in mind that if `new_value.is_some()` (you probably want `if let Some(nv) = new_value` btw), then you do _not_ want to unlink the node, and you also do _not_ want to free the `BinEntry`.

tests/basic.rs
```diff
@@ -37,6 +37,42 @@ fn get_key_value_empty() {
...
+}
+
+#[test]
+fn replace_old_value_non_matching() {
+    let map = HashMap::<usize, usize>::new();
+    {
+        let guard = epoch::pin();
+        map.insert(42, 42, &guard);
+        // TODO: fit test with an actual non valid old_value
+        let old = map.replace_node(&42, None, None, &guard);
```
danielSanchezQ: @jonhoo , How could I create a testing `Shared<V>` here?, Also, how can I actually retrieve that from the map?
Sorry for the mess, I'm just dealing with this crossbeam pointers for the first time.

src/map.rs
```diff
@@ -1091,6 +1117,23 @@ where
...
+    {
+        let guard = epoch::pin();
+        // removed selected keys
+        for (k, v) in self.iter(&guard) {
+            if !f(k, v) {
+                // TODO: right now this removing the element by key without checking the value. The value should be used somehow in old_value
+                self.replace_node(k, None, None, &guard);
```
danielSanchezQ: @jonhoo , Do we really need to check the old value here as in the java code does? I think I do not really understand that behavior:
If the value has change then why would I lose interest in removing it? I think if some thread commands to remove an entry should be ok, doesn't matter the value that entry holds. Or not?

src/map.rs
```diff
@@ -1025,28 +1045,34 @@ where
...
-                                unsafe { pred.deref() }
-                                    .as_node()
-                                    .unwrap()
-                                    .next
-                                    .store(next, Ordering::SeqCst);
-                            } else {
-                                // or by setting the next node as the first BinEntry if there is no previous entry
-                                t.store_bin(i, next);
+                            // found the node but we have a new value to replace the old one
+                            if let Some(nv) = new_value {
```
jonhoo: This should also only happen if the `old_value` matches (if it is `Some`)

src/map.rs
```diff
@@ -1025,28 +1045,34 @@ where
...
+                                n.value.store(Owned::new(nv), Ordering::SeqCst);
+                                break;
+                            // just remove the node if the value is the one we expected at method call
+                            } else if old_value.is_none() || old_value.unwrap() == ev {
```
```suggestion
                            } else if old_value.map(|ov| ov == ev).unwrap_or(true) {
```

src/map.rs
jonhoo: I believe the thinking is that if the old value has changed, then who are we to say that the user still wants to remove the element? I suppose we could also provide a method like `retain_force` (maybe pick a better name) that removes the value even if it has changed.

tests/basic.rs
jonhoo: I think you can just do
```rust
let old_value = Shared::from(map.get(&42, &guard).unwrap());
```
Since `replace_node` is a private function though, this probably has to be a unit-test directly in `map.rs`.

src/map.rs
danielSanchezQ: Actually here [this](https://github.com/jonhoo/flurry/pull/36#discussion_r371968383) is needed also because of [this](https://github.com/jonhoo/flurry/blob/0c3bfb617be08279aaf54ddbe15cf22a9bed69ee/jsr166/src/ConcurrentHashMap.java#L1630), right?
So, the value here I have it as a `&V` already, but actually a [`Shared<'g, V>`](https://github.com/jonhoo/flurry/pull/36/files#diff-178822c448ce6afe0e383be6f514626eL1005) is what is used inside the replace method.
`Shared::from(self.get(k, &guard).unwrap())` makes the compiler cry :/

tests/basic.rs
danielSanchezQ: I cannot create the pointer like that. Maybe I'm using the wrong type?
Same issue [here](https://github.com/jonhoo/flurry/pull/36#discussion_r372006985)

src/map.rs
```diff
@@ -945,6 +945,26 @@ where
...
+    /// Replaces node value with v, conditional upon match of cv.
+    /// If resulting value does not exists it removes the key (and its corresponding value) from this map.
+    /// This method does nothing if the key is not in the map.
+    /// Returns the previous value associated with the given key.
+    ///
+    /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
+    /// form must match those for the key type.
+    pub fn replace_node<'g, Q>(
+        &'g self,
+        key: &Q,
```
jonathanGB: IMHO, the signature of this function is getting complicated with `new_value` and `old_value`(trying to remember which of the *old* and the *new* comes first), especially considering this function is part of the public API. Should we embed that information in a struct, to make it clearer? Say
```rust
struct NewOldValue {
  new: Option<V>,
  old: Option<Shared<'g, V>>,
}

pub fn replace_node<'g, Q>(&'g self, key: &Q, values: NewOldValue, guard: &'g Guard) -> ...
```

src/map.rs
```diff
@@ -945,6 +945,26 @@ where
...
     pub fn remove<'g, Q>(&'g self, key: &Q, guard: &'g Guard) -> Option<&'g V>
+        where
+            K: Borrow<Q>,
+            Q: ?Sized + Hash + Eq,
+    {
+        self.replace_node(key, None, None, guard)
+    }
+
+    /// Replaces node value with v, conditional upon match of cv.
+    /// If resulting value does not exists it removes the key (and its corresponding value) from this map.
```
jonathanGB: nit

```suggestion
    /// If resulting value does not exist, it removes the key (and its corresponding value) from this map.
```

src/map.rs
danielSanchezQ: `replace_node` should not be public actually and will be change during the PR. Sorry for that.
This said, I'm down for that signature also.

src/map.rs
danielSanchezQ: Thanks :)

src/map.rs
jonhoo: What is the actual error you get? Does `Shared::from(... as *const _)` work?

src/map.rs
jonhoo: I would like it to match the Java version pretty closely so that future changes are easy to adopt to. I think we may want to call it `observed_value` or something like that to communicate intent over just "old".

src/map.rs
danielSanchezQ: Sorry, should have report the error before:
`v` is of type `&V` already, taken from the iterator, so, when trying the "casting" into the `*const` I bump into this lifetime issue:
```
error[E0309]: the parameter type `V` may not live long enough
    --> src/map.rs:1134:33
     |
150  | impl<K, V, S> HashMap<K, V, S>
     |         - help: consider adding an explicit lifetime bound `V: 'g`...
...
1134 |                 let old_value : Shared<'g, V> = Shared::from(v as *const _);
     |                                 ^^^^^^^^^^^^^
     |
note: ...so that the type `V` will meet its required lifetime bounds
    --> src/map.rs:1134:33
     |
1134 |                 let old_value : Shared<'g, V> = Shared::from(v as *const _);
     |                                 ^^^^^^^^^^^^^

```

jonhoo: Hmm, I thought the bound `V: 'g` would be inferred in the response from the iterator. It returns `&'g V`, right? Can you try being more explicit with `Shared::from(v as *const V)`?

danielSanchezQ: I found a way of making it compile:

```rust
impl<'g, K, V, S> HashMap<K, V, S>
where
    K: Sync + Send + Clone + Hash + Eq,
    V: Sync + Send + 'g,
    S: BuildHasher,
{

    /// Retains only the elements specified by the predicate.
    ///
    /// In other words, remove all pairs (k, v) such that f(&k,&v) returns false.
    pub fn retain<F>(&mut self, mut f: F)
        where
            F : FnMut(&K, &V) -> bool
    {
        let guard = epoch::pin();
        // removed selected keys
        for (k, v) in self.iter(&guard) {
            if !f(k, v) {
                // TODO: right now this removing the element by key without checking the value. The value should be used somehow in old_value
                let old_value : Shared<'g, V> = Shared::from(v as *const V);
                self.replace_node(
                    k,
                    None,
                    Some(old_value),
                    &guard,
                );
            }
        }
    }
}
```
I made and implicit `impl` for that live time, that solves the life time issues. The compiler error was actually saying so.

jonhoo: Ohhh, wait, you actually had `Shared<'g, V>` in the code. Yeah, that probably won't work. I think what you want is just:
```rust
let old_value = Shared::from(v as *const V);
```
or _maybe_
```rust
let old_value: Shared<'_, V> = Shared::from(v as *const V);
```
The `'g` you've added to the `impl` block now is super weird, and I'm not sure how it even compiles since `'g` isn't bound to anything. When I wrote `'g`, what I mean was the lifetime of the `guard` that you create _inside_ the function.

danielSanchezQ: Aaaahhh, that was indeed the issue :/
I was using the specific lifetime and it wasn't even declared in the block.
Thanks! ⚙️

danielSanchezQ:
I added `retain_force`, similar to `retain` but do not check on values and just enforces entry removal.

src/map.rs
```diff
@@ -1152,28 +1173,36 @@ where
...
+                                } else {
+                                    // or by setting the next node as the first BinEntry if there is no previous entry
+                                    t.store_bin(i, next);
+                                }
+
+                                // in either case, mark the BinEntry as garbage, since it was just removed
+                                // safety: as for val below / in put
+                                unsafe { guard.defer_destroy(e) };
+
+                                // since the key was found and only one node exists per key, we can break here
```
jonhoo: I think you want to `break` outside of the `if observed_value`, since we _did_ find the key, so there's no use searching further.

src/map.rs
```diff
@@ -1152,28 +1173,36 @@ where
...
-                            unsafe { guard.defer_destroy(e) };
-
-                            // since the key was found and only one node exists per key, we can break here
-                            break;
+                            // just remove the node if the value is the one we expected at method call
+                            if observed_value.map(|ov| ov == ev).unwrap_or(true) {
+                                // found the node but we have a new value to replace the old one
+                                if let Some(nv) = new_value {
+                                    n.value.store(Owned::new(nv), Ordering::SeqCst);
+                                    break;
```
jonhoo: Maybe add a comment here saying that we do not want to remove if we are replacing the value, and that's why we `break`?

src/map.rs
```diff
@@ -1152,28 +1173,36 @@ where
...
+                                }
+                                old_val = Some(ev);
```
jonhoo: Maybe a comment saying "we remember the old value so that we can return it and mark it for deletion below".

src/map.rs
```diff
@@ -1218,6 +1247,44 @@ where
         None
     }
 
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs (k, v) such that f(&k,&v) returns false.
+    ///
+    /// This method relies on [`HashMap::replace_node`] and do not force entries deletion if the
```
jonhoo: Since `replace_node` is not public, I don't think it should be mentioned in the docs. How about:
```rust
/// If `f` returns `false` for a given key/value pair, but the value for that pair is concurrently modified before the removal takes place, the entry will not be removed. If you want the removal to happen even in the case of concurrent modification, use [`HashMap::retain_force`].

File: src/map.rs [hunk 12]
```diff
@@ -1218,6 +1247,44 @@ where
...
+                self.replace_node(k, None, Some(old_value), &guard);
+            }
+        }
+    }
+
+    /// Retains only the elements specified by the predicate.
+    ///
+    /// In other words, remove all pairs (k, v) such that f(&k,&v) returns false.
+    ///
+    /// This method force entry deletion even if if it is updated concurrently.
```
```suggestion
    /// This method always deletes any key/value pair that `f` returns `false` for, even if if the value is updated concurrently. If you do not want that behavior, use [`HashMap::retain`].
```

src/map.rs
```diff
@@ -1218,6 +1247,44 @@ where
...
+    pub fn retain_force<F>(&mut self, mut f: F)
+    where
+        F: FnMut(&K, &V) -> bool,
+    {
+        let guard = epoch::pin();
+        // removed selected keys
+        for (k, v) in self.iter(&guard) {
+            if !f(k, v) {
+                self.remove(k, &guard);
```
```suggestion
                self.replace_node(k, None, None, &guard);
```

jonhoo: Just for consistency.

src/map.rs
danielSanchezQ: Indeed, it was misplaced. Good catch! Thanks!

src/map.rs
danielSanchezQ: Absolutely :)

tests/basic.rs
```diff
@@ -445,3 +445,46 @@ fn from_iter_empty() {
...
+    let mut map: HashMap<usize, usize> = (0..size).map(|x| (x, x)).collect();
+    map.retain(|_, _| true);
+    assert_eq!(map.len(), size);
+}
+
+#[test]
+fn retain_some() {
+    let mut map: HashMap<u32, u32> = (0..10).map(|x| (x, x)).collect();
+    map.retain(|_, v| *v >= 5);
+    assert_eq!(map.len(), 5);
```
jonhoo: We should check that the right elements are in the map here I think. Maybe just create a new map with the kept elements and then assert that the two maps are equal?

tests/basic.rs
```diff
@@ -445,3 +445,46 @@ fn from_iter_empty() {
...
+    let mut map = HashMap::<&'static str, u32>::new();
+    map.retain_force(|_, _| false);
+    assert_eq!(map.len(), 0);
+}
+
+#[test]
+fn retain_force_some() {
+    let mut map: HashMap<u32, u32> = (0..10).map(|x| (x, x)).collect();
+    map.retain_force(|_, v| *v >= 5);
+    assert_eq!(map.len(), 5);
```
jonhoo: Same here as above, let's check that the right elements are kept.

tests/basic.rs
danielSanchezQ: Yep, that is a much better way of making the test.

jonhoo:
Excellent, thank you!

danielSanchezQ:
Thank you for your patience!!

jonhoo:
Now let's see if this change passes the new miri and sanitizer CI tests :o
https://github.com/jonhoo/flurry/pull/37

PR #37, opened 2020-01-27
Add miri CI test

jonhoo:
Currently blocked on https://github.com/rust-lang/miri/pull/1150 landing on nightly and on https://github.com/crossbeam-rs/crossbeam/pull/458.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,16 @@ azure-devops = { project = "jonhoo/jonhoo", pipeline = "flurry", build = "15" }
 codecov = { repository = "jonhoo/flurry", branch = "master", service = "github" }
 maintenance = { status = "experimental" }
 
+[features]
+sanitize = ['crossbeam-epoch/sanitize']
+
 [dependencies]
-crossbeam-epoch = "0.8"
+crossbeam-epoch = "0.9"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
 
 [dev-dependencies]
 rand = "0.7"
+
+[patch.crates-io]
+crossbeam-epoch = { git = "https://github.com/cynecx/crossbeam.git", branch = "fix-unsoundness" }
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,6 +29,45 @@ jobs:
        condition: ne(variables.CACHE_RESTORED, 'true')
      - script: cargo deny check
        displayName: cargo deny
+ - job: miri
+   displayName: "Run miri on test suite"
+   dependsOn: deny
+   continueOnError: true # since miri sometimes isn't on nightly
+   pool:
+     vmImage: ubuntu-16.04
+   steps:
+     - template: install-rust.yml@templates
+       parameters:
+         rust: nightly
+         components:
+           - miri
+     # ignore leaks due to https://github.com/crossbeam-rs/crossbeam/issues/464
+     - bash: yes | cargo miri -Zmiri-ignore-leaks test
+       displayName: cargo miri test
+ - job: asan
+   dependsOn: deny
+   displayName: "Run address sanitizer on test suite"
+   pool:
+     vmImage: ubuntu-16.04
+   steps:
+     - template: install-rust.yml@templates
+       parameters:
+         rust: nightly
+     - script: |
+           env ASAN_OPTIONS="detect_odr_violation=0" RUSTFLAGS="-Z sanitizer=address" cargo test --release --features sanitize
+       displayName: cargo -Z sanitizer=address test
+ - job: lsan
+   dependsOn: deny
+   displayName: "Run leak sanitizer on test suite"
+   pool:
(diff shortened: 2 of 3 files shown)


jonhoo:
@danielSanchezQ Interesting -- it looks like the tests from #36 found a memory leak. Want to take a stab at fixing them? The logs are [here](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=529&view=logs&j=bc994297-029d-5157-6f73-4fe90c912368&t=f2b50a2b-32dd-5a67-fba7-e24e52af645e&l=93), and the relevant entries are:
```
Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #4 0x5609cbc139e2 in crossbeam_epoch::atomic::Owned$LT$T$GT$::new::hedac2cb2a6626c3b /home/vsts/.cargo/git/checkouts/crossbeam-308b9896d3fb1a0a/483d347/crossbeam-epoch/src/atomic.rs:664:19
    #5 0x5609cbc00671 in flurry::map::HashMap$LT$K$C$V$C$S$GT$::put::h8a55c48053512f17 /home/vsts/work/1/s/src/map.rs:362:19
    #6 0x5609cbbfff04 in flurry::map::HashMap$LT$K$C$V$C$S$GT$::insert::hf7a9b82571be5615 /home/vsts/work/1/s/src/map.rs:346:8
    #7 0x5609cbc08bb9 in flurry::map::replace_existing_observed_value_matching::_$u7b$$u7b$closure$u7d$$u7d$::h79bf84ff713dd6f5 /home/vsts/work/1/s/src/map.rs:1710

Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #4 0x5609cbc139e2 in crossbeam_epoch::atomic::Owned$LT$T$GT$::new::hedac2cb2a6626c3b /home/vsts/.cargo/git/checkouts/crossbeam-308b9896d3fb1a0a/483d347/crossbeam-epoch/src/atomic.rs:664:19
    #5 0x5609cbc00671 in flurry::map::HashMap$LT$K$C$V$C$S$GT$::put::h8a55c48053512f17 /home/vsts/work/1/s/src/map.rs:362:19
    #6 0x5609cbbfff04 in flurry::map::HashMap$LT$K$C$V$C$S$GT$::insert::hf7a9b82571be5615 /home/vsts/work/1/s/src/map.rs:346:8
    #7 0x5609cbc08b99 in flurry::map::replace_existing::_$u7b$$u7b$closure$u7d$$u7d$::h9f40593e19e439cb /home/vsts/work/1/s/src/map.rs:1698
```
Specifically, it looks like the value inserted here for `replace_existing`
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/src/map.rs#L1695
and here for `replace_existing_observed_value_matching`
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/src/map.rs#L1707
are not freed when we later replace them. I'm not entirely sure why, but some prints may help clarify the situation?

danielSanchezQ:
Sure, I’ll investigate the issue!

danielSanchezQ:
This piece of code solves the memory leak:

```rust
                                if let Some(nv) = new_value {
                                    let dropable_old = n.value.load(Ordering::SeqCst, guard);
                                    n.value.store(Owned::new(nv), Ordering::SeqCst);
                                    // we are just replacing entry value and we do not want to remove the node
                                    // so we stop iterating here
                                    drop(unsafe {dropable_old.into_owned()});
                                    break;
                                }
```
I'm forcing the drop. Probably is not the best way, but it gives some hints.
I am not sure why the resource is not being dropped, where could we have another shared pointer to that?

jonhoo:
Ah, no, I see what's going on! The `break` inside that `if` means that
```rust
old_value = Some(ev);
```
does not get executed, which in turn means that the deferred destroy further down does not get executed
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/src/map.rs#L1238
I _think_ this would be fixed by moving
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/src/map.rs#L1187
up to just above the if. Will implement that!

jonhoo:
In fact, thinking some more on it, I think this assertion is wrong:
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/src/map.rs#L1697
It should be `Some` since we replaced an old value! The fix above also fixes that.

jonhoo:
That worked :tada:

danielSanchezQ:
Aaaah, I see it now. Actually it was my fault there. I understood that since we are not really removing the entry nothing should be returned. But ofc it makes sense that the old value is returned. Sorry for that!

Issue #34, opened 2020-01-26
Use loom in the tests


ZenTauro:
We could use [loom](https://github.com/tokio-rs/loom) to test multiple permutations in the tests.

ZenTauro:
I've been working on it, right now all the test pass except the `concurrent_insert` and `concurrent_remove` https://github.com/ZenTauro/flurry/commit/fdb2bf26c79d4ae7d4ad9999d7f7b106d223d08c

jonhoo:
I would love for us to also run the tests through miri!

jonhoo:
See also https://github.com/jonhoo/flurry/pull/37

ZenTauro:
Nice, however, when I run the test with miri and loom they don't produce
any output.

jonhoo:
Interesting.. What _do_ they produce?

ZenTauro:
It just exits with 0

jonhoo:
Hmm, I'm not sure -- @carllerche?

Separately, I do not think we want to use `debug_assertions` to determine whether or not to use `loom`. Instead, `loom` tests should probably be separated out and written explicitly for that purpose given how slow they can be. We also will want to run `loom` tests in release mode, otherwise they'll be far too slow. In tokio I know they require `env "RUSTFLAGS=--cfg loom"` and then guard the loom tests with `#[cfg(loom)]`.

carllerche:
I don't know anything about miri + loom :) seems rough to mix the two. I don't think miri supports threading.

Also, correct re: `RUSTFLAGS="--cfg loom"`.

Using loom right now is not friendly, if anyone is so inclined, we could definitely use some UX help :)

jonhoo:
Ah, @ZenTauro, were you trying to run both _at the same time_? Yeah, that probably won't work. I think we probably want to run the two independently.

ZenTauro:
> Ah, @ZenTauro, were you trying to run both _at the same time_? Yeah, that probably won't work. I think we probably want to run the two independently.

Yeah, I didn't know they were interfering

> Also, correct re: `RUSTFLAGS="--cfg loom"`.

I'll put them behind `--cfg loom`

ZenTauro:
I am trying to implement a macro that can annotate tests in order to run them with loom only if configured to do so. I am not really familiar with proc macros so I haven't been able to make a lot of progress, if someone is interested I'll appreciate some help in the implementation.
[...]

PR #39, opened 2020-01-29
Add test which attempts to encounter a concurrent resize and retrieval

jhinch:
I noticed that one of the areas of the code which had low test coverage was the [BinEntry::Moved branch in node.rs](https://github.com/jonhoo/flurry/blob/master/src/node.rs#L107). This test is designed to exercise that codepath. At least on my machine it seems to exercise the codepath but it is non-deterministic whether it will actually encounter the path.

### DIFF
--- a/src/node.rs
+++ b/src/node.rs
@@ -162,6 +162,16 @@ mod tests {
         }
     }
 
+    fn drop_entry(entry: BinEntry<usize, usize>) {
+        // currently bins don't handle dropping their
+        // own values the Table is responsible. This
+        // makes use of the tables implementation for
+        // convenience in the unit test
+        let mut table = Table::<usize, usize>::new(1);
+        table.store_bin(0, Owned::new(entry));
+        table.drop_bins();
+    }
+
     #[test]
     fn find_node_no_match() {
         let guard = &crossbeam_epoch::pin();
@@ -171,6 +181,7 @@ mod tests {
         node1.next.store(Owned::new(entry2), Ordering::SeqCst);
         let entry1 = BinEntry::Node(node1);
         assert!(entry1.find(1, &0, guard).is_null());
+        drop_entry(entry1);
     }
 
     #[test]
@@ -184,6 +195,7 @@ mod tests {
                 .key,
             2
         );
+        drop_entry(entry);
     }
 
     #[test]
@@ -201,6 +213,7 @@ mod tests {
                 .key,
             5
         );
+        drop_entry(entry1);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


jhinch:
Of course it decides to not hit the codepath in CI. How would you feel about me adding a unit test for the `find` method so it can be tested deterministically

jonhoo:
A unit test for that would be fantastic!

jonhoo:
I also wonder whether you might be able to trigger more resizes by forcing all the keys into a single bucket like we do here:
https://github.com/jonhoo/flurry/blob/5fe084b408eebd984fec28ac5a8888ea0ffeab5f/tests/basic.rs#L88-L104

jhinch:
At least when testing locally, using the one bucket hasher makes the `BinEntry::Moved` be encountered less often. I believe this is because the transfer function will transfer entire runs of entries when it can, so you actually want quite shallow bins to trigger the desired behaviour

jonhoo:
This looks great! I do still wish we had a reliable way to check that the `next_table` pointer _in_ `Moved` was set correctly, which basically requires a resize to take place while we run a `find`. Maybe one way to do that is to set up a very large table in the test, then call `reserve` with a value large enough to trigger a resize in a separate thread, and then immediately do `get`s on the main thread?

jhinch:
I would guess the only way to do it reliably is using a tool like loom to test each of the permutations

jonhoo:
I _think_ the strategy I outlined above should actually work pretty reliably, but yes, ultimately we'll want something like loom as well (#34).

jonhoo:
@jhinch would you be interested in also trying to implement the test strategy above?

jhinch:
Apologies. I thought your comments were more speculative of how things _could_ be done rather than of something which should be changed as part of this PR. I have switched it over to use `.reserve()` instead and increased it to resize to 32K to increase likelihood of a resize being triggered

jonhoo:
No worries. It was a bit of a speculative comments :p

I think you'll _first_ want to do many many inserts, and _then_ start the thread that does resizing. Otherwise the resizes will all finish very quickly, and you may not even see it.

jhinch:
It looks like I've introduced a memory leak in the unit tests I added. I'm working through the problem now.

jonhoo:
Ah, yes, the tests you've written never actually free any of the memory they allocate since you do not create a `Table` at any point (which is the thing that actually calls `defer_destroy` on stuff in its bins).

jhinch:
Alright, all tests are passing again and merge conflicts have been resolved

jonhoo:
Great, thanks!

PR #40, opened 2020-01-29
Fix broken doclinks which are causing the docs to fail to build.

jhinch:
Currently, docs.rs is broken for flurry: https://docs.rs/crate/flurry/0.0.2/builds/215675. This PR fixes it.

I think it would probably be a good idea to fail the build if doclinks are broken. However, I'm not familiar with azure-pipelines, so I'm not entirely sure how to set this up.

### DIFF
--- a/src/iter/iter.rs
+++ b/src/iter/iter.rs
@@ -4,7 +4,7 @@ use std::sync::atomic::Ordering;
 
 /// An iterator over a map's entries.
 ///
-/// See [`HashMap::iter`] for details.
+/// See [`HashMap::iter`](crate::HashMap::iter) for details.
 #[derive(Debug)]
 pub struct Iter<'g, K, V> {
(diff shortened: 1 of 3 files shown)


jonhoo:
Oh, how weird! I thought
https://github.com/jonhoo/flurry/blob/1725d8837fecc1b78fd1ac12d97082ee1430f2e2/src/lib.rs#L199
should take care of that, but apparently not. It looks like the [CI script](https://github.com/crate-ci/azure-pipelines/blob/2770896d2dc7b3df4493f436c7004a9caeb02172/default.yml) we are using runs `cargo check` and `cargo test`, but not `cargo doc`, and only the latter catches this error! I'm pushing https://github.com/crate-ci/azure-pipelines/commit/759793329d3451468024686f3c7640331c12eddf now to fix that.

PR #42, opened 2020-01-29
Implement clear method

Stupremee:
This pr adds the clear method to FlurryHashMap.


Stupremee:
I need some help.
Currently, the code doesn't compile but that's not the problem.
How can I port the assignment in line 1173 of the java implementation to Rust?

jonhoo:
Let's see..
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/jsr166/src/ConcurrentHashMap.java#L1173-L1175
is really
```java
Node<K,V> p = null;
if (fh >= 0) {
  p = f;
} else if (f instanceof TreeBin) {
  p = ((TreeBin<K,V>)f).first;
}
```
Recall that the Java code uses a negative hash code to indicate that a node is of a "special type" (like `TreeBin` or `ReservationNode`). Since we do not handle `TreeBin`, what this is really saying is "set `p` to be a pointer to the first `BinEntry` assuming that entry is a `BinEntry::Node`". The code then walks the linked list (`while (p != null)`) if the bin was indeed headed by a `BinEntry::Node`, and then sets the bin to `null` regardless of whether `p` was non-null or not
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/jsr166/src/ConcurrentHashMap.java#L1176-L1180

domenicquirl:
Dang, beat me to it 😅

Stupremee:
The PR is almost ready (I guess).
There are only 3 things I need help with.
1. Which lock should I use for the synchronization? (line 491)
2. Should the clear method do something special if the table isn't allocated? (line 470)
3. What should happen in in line 489?

domenicquirl:
> 1. Which lock should I use for the synchronization? (line 491)
 If you match against `BinEntry::Node(ref head)` (instead of `Node(_)`), you can use `let head_lock = head.lock.lock();` and then do `drop(head_lock);` where the `synchronized` ends (the lock is also dropped automatically, but up to this point we usually stated this explicitly).
> 2. Should the clear method do something special if the table isn't allocated? (line 470)
As far as I can tell, the Java code just skips over the `while` loop and does not execute the following `if`, so I wouldn't worry about that. Other methods handle this similarly, and allocating the table is done by methods that insert into the table such as `putVal` or `compute`.
> 3. What should happen in in line 489?
I'm wondering about the line above at the moment, it seems you `deref()` `s` only if it `is_null()`? For the actual `null` case, the only time returns a `null` value is if the table is already `null` (i.e. not yet allocated). In this case you can just `break` (the Java case finishes the `else` and then fails the `null` check in the `while` loop). Maybe you can also `return`, but that depends on if `delta` can possibly be modified before reaching this point. 

I'd definitely suggest adding a test for clearing an empty map if you don't have one already.

Stupremee:
Yea I'm already working on it.

Stupremee:
The test passes :+1:

src/map.rs
```diff
@@ -346,6 +346,52 @@ where
...
+        let mut idx = 0usize;
+
+        let mut table = self.table.load(Ordering::SeqCst, guard);
+
+        // Safety: self.table is a valid pointer because we checked it above.
+        while !table.is_null() && idx < unsafe { table.deref() }.len() {
+            let tab = unsafe { table.deref() };
+            let node = tab.bin(idx, guard);
+            if node.is_null() {
+                idx = idx + 1;
```
```suggestion
                idx += 1;
```

src/map.rs
```diff
@@ -346,6 +346,52 @@ where
...
+                continue;
+            }
+            // Safety: node is a valid pointer because we checked
+            // it in the above if stmt.
+            let node = unsafe { node.deref() };
+            match node {
+                BinEntry::Moved(next_table) => {
+                    table = self.help_transfer(table, *next_table, guard);
+                    idx = 0;
```
```suggestion
                    // start from the first bin again following the resize
                    idx = 0;
```

src/map.rs
```diff
@@ -346,6 +346,52 @@ where
...
+                }
+                BinEntry::Node(ref node) => {
+                    let lock = node.lock.lock();
```
jonhoo: We've called this `head_lock` in other parts of the code since you do not take the lock of _every_ node, just the first one. Let's continue to do that here :)

jonhoo: We need to re-check that this is still the head of the bin after taking the lock
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/jsr166/src/ConcurrentHashMap.java#L1172
You can see how we do that in other parts of the code here for example:
https://github.com/jonhoo/flurry/blob/9544337e6d9ce5a4d827dd7af63cdb4a5ae24bd9/src/map.rs#L446-L451

src/map.rs
```diff
@@ -346,6 +346,52 @@ where
...
+                    let mut p: Option<&Node<K, V>> = Some(node);
+                    while p.is_some() {
+                        delta = delta - 1;
```
```suggestion
                        delta -= 1;
```

src/map.rs
```diff
@@ -346,6 +346,52 @@ where
...
+                        p = match p.unwrap().next.load(Ordering::SeqCst, guard) {
+                            s if s.is_null() => None,
+                            s => unsafe { s.deref() }.as_node(),
```
jonhoo: Let us make this:
```rust
Some(unsafe { s.deref() }.as_node().expect("Node.next should always be BinEntry::Node"))
```

src/map.rs
```diff
@@ -346,6 +346,52 @@ where
...
+                        };
+                    }
+                    tab.store_bin(idx, Shared::null());
+                    idx = idx + 1;
```
```suggestion
                    idx += 1;
```

tests/basic.rs
```diff
@@ -7,6 +7,21 @@ fn new() {
...
+    let guard = epoch::pin();
+    {
+        map.insert(0, 1, &guard);
+        map.insert(1, 1, &guard);
+        map.insert(2, 1, &guard);
+        map.insert(3, 1, &guard);
+        map.insert(4, 1, &guard);
+    }
+    map.clear(&guard);
+    assert!(map.len() == 0);
```
jonhoo:
```suggestion
    assert!(map.is_empty());
```

jonhoo:
```suggestion
    assert!(map.is_empty());
```

jonhoo:
This is looking good! Could you merge `master` just to check the latest CI, and then we'll merge?

src/map.rs
```diff
@@ -346,6 +346,61 @@ where
         self.put(key, value, false, guard)
     }
 
+    /// Removes all entries from this map.
+    pub fn clear<'g>(&'g self, guard: &'g Guard) {
```
jonhoo: This does not need `'g` since we're not returning any references.

jonhoo:
Interesting.. So, it looks like there is a [memory leak](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=560&view=logs&j=bc994297-029d-5157-6f73-4fe90c912368&t=f2b50a2b-32dd-5a67-fba7-e24e52af645e&l=133) where values that are inserted in your test are not actually freed when `clear` is called.

Looking at the code, the issue is actually pretty straightforward: you never call `defer` destroy on the bins or values you drop! Inside this loop
https://github.com/jonhoo/flurry/blob/42cc12cfa4e3dc88acd4565aef858434a87648b4/src/map.rs#L381
You need to do
```rust
guard.defer_destroy(p.value.load());
```
You _also_ need to `defer_destroy` all of the `BinEntry` that you walk over. In the Java code, this happens automatically through the garbage collector, but not for us. See
https://github.com/jonhoo/flurry/blob/42cc12cfa4e3dc88acd4565aef858434a87648b4/src/map.rs#L1260
and
https://github.com/jonhoo/flurry/blob/42cc12cfa4e3dc88acd4565aef858434a87648b4/src/map.rs#L1294

Stupremee:
Ah okay. I will implement it asap

src/map.rs
```diff
@@ -380,17 +380,23 @@ where
                     let mut p: Option<&Node<K, V>> = Some(node);
                     while p.is_some() {
                         delta -= 1;
+                        unsafe {
```
jonhoo: This should include a safety argument -- see `remove`.

src/map.rs
```diff
@@ -380,17 +380,23 @@ where
...
                                 unsafe { s.deref() }
                                     .as_node()
                                     .expect("Node.next should always be BinEntry::Node"),
                             ),
-                        }
+                        };
                     }
                     tab.store_bin(idx, Shared::null());
                     idx += 1;
+                    unsafe {
```
jonhoo: This will only free the _first_ node, but _all_ the nodes need to be freed.
Also, this should include a safety argument -- see `remove`.

Stupremee: But `raw_node' is the node to which the current index points which increases every loop cycle. Doesn't this mean that all nodes are destroyed?

jonhoo: Ah, but remember that every bin is a linked list. `raw_node` is the _first_ node in the linked list in bin `idx`, but you also need to free all the nodes along the `.next` pointers (that's what the `while` loop just above walks).

Stupremee: Ah okay, makes sense.

src/map.rs
Stupremee: Can I just copy the safety note from the `replace_node` method?

Stupremee:
I'm not sure if the nodes and values are destroyed correctly.
Someone has to check it.

jonhoo:
Okay, so, having read through this code, I think the structure needs to change a little for the safety argument to be easy to express. I think the code we want in the `BinEntry::Node` case is this:
```rust
let head_lock = node.lock.lock();
// need to check that this is _still_ the head
let mut current_head = tab.bin(idx, guard);
if current_head != raw_node {
    // nope -- try the bin again
    continue;
}
// we now own the bin
// unlink it from the map to prevent others from entering it
tab.store_bin(idx, Shared::null());
// next, walk the nodes of the bin and free the nodes and their values as we go
// note that we do not free the head node yet, since we're holding the lock it contains
let mut p = node.next.load(Shared::null(), Ordering::SeqCst, guard);
while !p.is_null() {
    delta -= 1;
    p = {
        // safety: we loaded p under guard, and guard is still pinned, so p has not been dropped.
        let node = unsafe { p.deref() }.as_node().expect("entry following Node should always be a Node");
        let next = node.next.load(Shared::null(), Ordering::SeqCst, guard);
        let value = node.value.load(Ordering::SeqCst, guard);
        drop(node);

        // free the node's value
        // safety: any thread that sees this p's value must have read the bin before we stored null
        // into it above. it must also have pinned the epoch before that time. therefore, the
        // defer_destroy below won't be executed until that thread's guard is dropped, at which
        // point it holds no outstanding references to the value anyway.
        unsafe { guard.defer_destroy(value) };
        // free the bin entry itself
        // safety: same argument as for value above.
        unsafe { guard.defer_destroy(p) };
        next
    };
}
drop(head_lock);
// finally, we can drop the head node and its value
let value = node.value.load(Ordering::SeqCst, guard);
drop(node);
// safety: same as the argument for being allowed to free the nodes beyond the head above
unsafe { guard.defer_destroy(value) };
unsafe { guard.defer_destroy(raw_node) };
idx += 1;
```

You'll also want to change
```rust
let node = unsafe { raw_node.deref() };
match node {
```
to
```rust
match unsafe { raw_node.deref() } {
```
to ensure we don't leave the invalid reference `node` around after destroying `raw_node`. @domenicquirl do you want to double-check my logic here?

Stupremee:
That looks much better.

jonhoo:
@Stupremee great! could you incorporate that change into the PR?

Stupremee:
Yep. I will implement it asap

Stupremee:
Everything should work now. I had to change a bit of your code because there was an error in it.
The removal of the head_node was not added to the `delta` which resulted in an invalid delta size and therefore the count wasn't set to zero.

The lint is failing because in line 503 a reference is dropped. This could be resolved by changing the `as_node` method to return an owned value instead of a reference. Should I change the `as_node`  method as well?

jonhoo:
Hmm, no, I don't think your change is actually a legal one, since it will now drop the head node while we're still holding the lock inside of it. It needs to happen exactly like written in my version I believe. The solution to the bad `delta` is just to add `delta -= 1;` to right after
```rust
unsafe { guard.defer_destroy(raw_node) };
```
near the bottom of my code :)

src/map.rs
```diff
@@ -472,49 +471,56 @@ where
...
-                    unsafe {
-                        guard.defer_destroy(raw_node);
-                    }
                     drop(head_lock);
+                    idx += 1;
                 }
             };
         }
 
+        println!("delta: {}", delta);
```
jonhoo: This probably shouldn't be there!

Stupremee: Yes I forgot that

Stupremee:
Ahh. Makes sense. Thank you

jonhoo:
Hmm, the style thing is annoying. I was doing that just to make sure we didn't leave any dangling references. You can just replace each of those `drop(node);` statements with
```rust
// NOTE: do not use the reference in `node` after this point!
```

Stupremee:
Then the PR should be ready if there's nothing left to do.

jonhoo:
Fantastic! I'll merge once CI passed :tada:

domenicquirl:
Since you pinged me; your safety arguments seem fine to me 👍 
For most cases it's probably even enough that we load the table under the guard.

domenicquirl:
The interactions of this with other operations are super interesting though. In particular, following a `Moved` and restarting makes use of moving bins from high to low. We should maybe add a comment or at least keep this in mind. For other methods, the `Moved` case simply follows a reference, but here it interacts with the `clear` functionality.

Issue #8, opened 2020-01-23
Add benchmarks


jonhoo:
We should port the single-core benchmarks [from `hashbrown`](https://github.com/rust-lang/hashbrown/blob/master/benches/bench.rs), and the concurrent ones [from `dashmap`](https://github.com/xacrimon/dashmap/tree/master/benches) to measure the performance of flurry!

domenicquirl:
How do you feel about the dependencies for this? `dashmap` uses `criterion` which, [according to their docs](https://docs.rs/criterion/0.3.1/criterion/macro.criterion_main.html) only works if one disables the default benchmarking harness generated by `test`. On the contrary, `hashbrown` uses `test` and its `Bencher`. (Also `dashmap` uses `rayon`, but since we have #14 this might be added anyway).

Since `test::bench` is, I believe, still feature-gated and nightly-only, it might be easier to rewrite the `hashbrown` macros to use criterion, but either way we should probably settle on one solution for all benchmarks.

I'll start with the `dashmap` benches once this is clear.

jonhoo:
I think `criterion` is the way to go!

domenicquirl:
Alright! I'll open a PR when I got some code!

domenicquirl:
This can now be closed as #43 is merged 😊

Issue #30, opened 2020-01-25
Implement and run a "serious" concurrent hash table benchmarking suite


jonhoo:
It is easy to write simple benchmarks for concurrent hash tables. But when these data-structures hit real world data is when we learn how they truly perform. There has been much research on concurrent hash tables, and how to benchmark them, so let's build a benchmark (or set of benchmarks rather) inspired by that work! We can then run that benchmark against the various concurrent map implementations out there in Rust world, and (hopefully) get some useful data out of them.

I'm hoping this thread can act as a staging ground for designing this benchmark, and that it can then be forked off into its own stand-alone project.

Work of note to get started (please let me know if you know of others):

 - https://www.aimlab.org/haochen/papers/fgcs18-hash.pdf (§3.3)
 - https://arxiv.org/pdf/1601.04017.pdf (§8.3)
 - The [libcuckoo universal benchmark](https://github.com/efficient/libcuckoo/tree/master/tests/universal-benchmark)
 - https://www.usenix.org/legacy/event/atc11/tech/final_files/Triplett.pdf (§6.1)
 - [redis-benchmark](https://redis.io/topics/benchmarks)

/cc @xacrimon

xacrimon:
Ah, I would love to help but I am a tad busy right now.

jonhoo:
@xacrimon I CC'ed you mostly so that you could subscribe to this issue if you were interested, not in expectation that you'd necessarily do any work :)

jonhoo:
Having dug into this a bit more now, I think the `libcuckoo`'s [universal benchmark](https://github.com/efficient/libcuckoo/tree/master/tests/universal-benchmark) is the right starting point. It does _most_ of what we want, and the various papers add relatively little to what the benchmark already does. The one thing it is missing is concurrent access to the same key: each thread generates its own random key sequence, and in the 64-bit space, there are unlikely to be _any_ overlaps. We'll still see contention on _buckets_, which is arguably most important, but it'd be good to also measure same-key contention. But that I think we can leave for later.

jonhoo:
I wrote a first draft of a more "serious" concurrent benchmark: https://github.com/jonhoo/bustle

PR #43, opened 2020-01-29
Add benchmarks from dashmap and hashbrown

domenicquirl:
This addresses #8. The added files have all the original benchmarks from `dashmap` and `hashbrown` to compare against (this includes several other hashmap implementations), as well as implementations of the `dashmap`- and `hashbrown`-specific tests for `flurry`. I also tried to add an implementation of the `dashmap` tests for `flurry` that doesn't `pin()` the epoch for every operation (the original implementation uses `rayon`'s `par_iter().for_each()`, which requires the executed closure to be `Send + Sync`), however this requires setting up the threads differently and I'm not sure how much overhead this introduces. So have this in mind when using this comparison.

The main reason this is a draft is that I wanted to make sure about the licenses. `MIT` license, as far as I understand, requires license and copyright notices to be present in derived works. Am I correct in that this is required? If so, where exactly do we add them?


jonhoo:
For licenses, I _think_ the right thing to do is to include the license file from the repositories you copied from into `benches/` for example, with names like `dashmap.LICENSE`.

It would also be good to include a short comment at the top of each copied file saying where it was copied from.

Separately, I don't know that we need to pull in the benchmarks for other maps _into_ here. `benches/chashmap` for example. I think the trick is going to be to just have `benches/dashmap_flurry.rs`, which is the dashmap benchmark _implemented for `flurry`_. When we eventually implement our own benchmark suits (like #8), then we will also implement that for other implementations, but I don't think there's an upside to copy-pasting existing benchmarks into here. Maybe just point users to them in the `README` and include instructions on how to run them there?

I would also like us to have _separate_ benchmark files for the flurry version of different benchmarks. I'm imagining `benches/hashbrown_like.rs` and `benches/dashmap_like.rs`.

domenicquirl:
Wouldn't we want to compare our results against the other hashmaps? Of course one can run their benchmarks separately, I just thought it would be nice to have them available without needing separate projects. For comparing performance across releases, one could still execute only the `flurry` benchmarks. Or do you want this comparison only for such tests as #30? Since the execution is exactly the same between original (`dashmap`/`hashbrown`) and `flurry` benchmark code, it felt natural to at least include these two as a reference.

Regarding the license, I will update the branch accordingly. I can also separate `flurry.rs` into two files 👍

jonhoo:
We _do_ want to compare against other hashmaps, and while it is _convenient_ to copy the benchmarks from there to here, I don't think it's worth it. Better to avoid keeping multiple copies, and instead just refer people to where the master copy of the benchmarks for the other backends live.

benches/flurry_hashbrown.rs
```diff
@@ -1,12 +1,16 @@
-use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
-use hashbrown::HashMap;
 
-/* HASHBROWN:
+/* Benchmarks from `hashbrown` (https://github.com/rust-lang/hashbrown):
```
jonhoo: > ... adapted to `flurry` for comparison.

benches/flurry_dashmap.rs
```diff
@@ -0,0 +1,180 @@
+/* Benchmarks from `dashmap` (https://github.com/xacrimon/dashmap):
```
jonhoo: > ... adapted to `flurry` for comparison.

benches/README.md
```diff
@@ -0,0 +1,4 @@
+## Benchmarks
+
+Currently, benchmarks following those of [`dashmap`](https://github.com/xacrimon/dashmap/tree/master/benches) and [`hashbrown`](https://github.com/rust-lang/hashbrown/blob/master/benches/bench.rs) are provided. To compare against other hashmap implementations, the benchmarks located in the respective repositories may be executed. Note that `flurry`, like `dashmap`, uses [`criterion`](https://docs.rs/criterion/0.3.1/criterion/) (and [`rayon`](https://docs.rs/rayon/1.3.0/rayon/) for parallel testing), while `hashbrown` uses [`test::bench`](https://doc.rust-lang.org/test/bench/index.html).
```
jonhoo: We should probably note here that _our_ port of the `hashbrown` benchmarks still uses `criterion`?

Also, line breaks :p

domenicquirl: Is this unclear from the last sentence?

jonhoo: Ah, interesting. No, I suppose it is clear, I just got confused about what "flurry hashbrown" benchmarks would then look like.

benches/README.md
```diff
@@ -0,0 +1,6 @@
+## Benchmarks
+
+Currently, benchmarks following those of [`dashmap`](https://github.com/xacrimon/dashmap/tree/master/benches) and [`hashbrown`](https://github.com/rust-lang/hashbrown/blob/master/benches/bench.rs) are provided. 
+To compare against other hashmap implementations, the benchmarks located in the respective repositories may be executed. 
+Note that `flurry`, like `dashmap`, uses [`criterion`](https://docs.rs/criterion/0.3.1/criterion/) (and [`rayon`](https://docs.rs/rayon/1.3.0/rayon/) for parallel testing), while `hashbrown` uses [`test::bench`](https://doc.rust-lang.org/test/bench/index.html).
+
```
jonhoo: It would be good to include instructions here on how to run the benchmarks. And maybe even _also_ how to run the various upstream benchmarks for comparison.

domenicquirl: I'd maybe add a comment about `criterion` building its own harness. Other than that, you're simply referring to `cargo bench` for the benchmarks inside the project?

For the other benchmarks I'm not familiar enough with the Rust community to know if there is a preferred or particularly clever way to do this - I'd naively set up a "benchmarking project" with all the different maps as dependencies and run the benchmark files? Or check out some respective repository and run `bench` on the project?

jonhoo: Ah, sorry, I just meant "what commands do I run?". Like:

    To run the flurry benchmarks, just run

    ```console
    $ cargo bench
    ```
    
    To run the original `dashmap` benchmarks:
    ```console
    $ git clone ...
    $ cd dashmap
    $ cargo bench
    ```
    
    To run the original `hashbrown` benchmarks:
    
    ... etc.

jonhoo:
Have you tried actually running the benchmarks? I'm curious what kind of numbers you get and how they compare?

domenicquirl:
They seem... slow? Inserting in particular is way slower than the other maps. Basically...

- **against hashbrown**, `insert` and `insert_erase` run in the order of on ms, compared to 10-20 us in `hashbrown`. `get` holds up reasonably well though, still ~20 us to ~5-6, but given the added concurrency this seems much better than `insert`. Both of these use one guard across all operations of one benchmark run and relatively consistent across input distributions.
- **against dashmap**, `dashmap` inserts in something between 1 ms with 1 thread and .8 ms for multiple threads, however what's probably more important is throughput. This for some reason is not in the report produced by `criterion`, but was on the order of 100 MElements/s. We compare with about 25-12 ms depending on thread count, which is so slow it takes several minutes to run any benchmark and puts throughput at something like 5, maybe 10 MElements/s. This is using 1 guard per map operation, which admittedly is not what you would do in a scenario such as the benchmark.
- **against dashmap, but with only one guard per thread** (the different setup), we `insert` in 14-11 ms, which is better, but still considerably slower than `dashmap`. Note that this does not compare directly due to the different setup of the threads.
- **get against dashmap** seems really good! `dashmap` gets 700-500 us of inserts, while we get 1.5 ms on one thread, 800us on 2 threads and go down to below 300 us on 8 threads, even with `pin()` for every call to `get`. With only one guard, this improves to 800 to a bit more than 200 us, depending on thread count.

So yeah, overall `get` is promising but `insert` is _much_ slower against both references :/
Other maps were similar and generally beat us in throughput, with `get` being close and `insert`... not being close.

domenicquirl:
It's worth noting that I ran this on my personal machine while doing other tasks, so this is not the cleanest setup. I don't this that would be responsible for differences of this magnitude.

Is there a Rust profiling tool which could tell which part of `insert` takes so much time? That would be an interesting question.

jonhoo:
Oh, that is very interesting indeed! Let's land this and then do some profiling.

For profiling, my recommendation is to add:
```toml
[profile.release]
debug = true
[profile.bench]
debug = true
```
to `Cargo.toml`, start benchmark, and then use `perf record -g --graph dwarf -p $(pgrep -af target/release)` to record some profiling information. Then [feed the results into a flamegraph](https://github.com/jonhoo/inferno#as-a-binary) and see what sticks out at you!

domenicquirl:
I might have to defer this to someone else. I'm on a Windows machine at the moment and will be a little busy the next few weeks anyways. Perhaps it's better to open an issue for this, so someone can really take a good look at this?

jonhoo:
Yes, that makes sense! Want to open an issue and include your preliminary results?

PR #44, opened 2020-01-29
Make the crate `no_std` + `alloc`

GarkGarcia:
This is far from complete, but I went ahead and replaced `std` for `core` in all of the easy places. I also created the `std` feature and configured all relevant functions to only be exposed when using it.

I'd like some input before I proceed.


src/map.rs
```diff
@@ -1,20 +1,19 @@
...
-    Once,
-};
+use std::sync::Once;
 
 macro_rules! isize_bits {
     () => {
-        std::mem::size_of::<isize>() * 8
+        core::mem::size_of::<isize>() * 8
     };
 }
```
GarkGarcia: Side note: since `core::mem::size_of` is a `const fn`, couldn't this be a `cons`?

GarkGarcia:
On regards to your point about having to figure out what to do with the [default value of `S`](https://github.com/jonhoo/flurry/blob/2a5bd258fa805e89715f7dbd52b63bddf736490f/src/lib.rs#L262), apparently the compiler is happy with `S = RandomState`, even though it is not defined in a `no_std` environment 🤷. I haven't tried compiling it with `no_std` though.

src/map.rs
jonhoo: Oh, you're right, it probably could (and should).

src/map.rs
```diff
@@ -1447,6 +1448,7 @@ where
     }
 }
 
+#[cfg(feature = "std")]
```
jonhoo: We could probably improve these various impls to be:
```rust
impl <K, V, S> FromIterator ... for HashMap<K, V, S>
where
    S: Default
```
That should cover `RandomState`, _and_ work in `no_std`!

src/map.rs
```diff
@@ -311,7 +312,7 @@ where
             let mut sc = self.size_ctl.load(Ordering::SeqCst);
             if sc < 0 {
                 // we lost the initialization race; just spin
-                std::thread::yield_now();
+                core::sync::atomic::spin_loop_hint();
```
jonhoo: I think we want to keep this `yield_now` if `std` is available.

jonhoo: Also, let's include the argument for why this is okay from #26

src/map.rs
```diff
@@ -95,6 +95,7 @@ where
     }
 }
 
+#[cfg(feature = "std")]
```
jonhoo: I wonder if we could play the same trick here of using `where S: Default`?

src/map.rs
```diff
@@ -85,6 +84,7 @@ pub struct HashMap<K, V, S = RandomState> {
     build_hasher: S,
 }
 
+#[cfg(feature = "std")]
```
jonhoo: Can we turn this into `where S: Default`?

src/map.rs
GarkGarcia: We can! The only reason we were locked to `RandomState` was that we needed a method for initializing the `BuildHasher`.

GarkGarcia: Done.

src/map.rs
GarkGarcia: Yep, good idea!

src/map.rs
GarkGarcia: Is there a way to execute code when a feature is not present? Something like `#[not(cfg(feature = "std"))]` ...

jonhoo: Almost!
```rust
#[cfg(not(feature = "std"))]
```

jonhoo:
I think we're still missing a couple of important things in `src/lib.rs`. [`crossbeam-epoch/src/lib.rs`](https://github.com/crossbeam-rs/crossbeam/blob/ebecb82c740a1b3d9d10f235387848f7e3fa9c68/crossbeam-epoch/src/lib.rs) is probably good inspiration. In particular, we'll need to add [`cfg_if`](https://github.com/alexcrichton/cfg-if) to `Cargo.toml`, and then something like
```rust
#![cfg_attr(not(feature = "std"), no_std)]

#[macro_use]
extern crate cfg_if;
#[cfg(feature = "std")]
extern crate core;

cfg_if! {
    if #[cfg(feature = "alloc")] {
        extern crate alloc;
    } else if #[cfg(feature = "std")] {
        extern crate std as alloc;
    }
}
```

src/map.rs
GarkGarcia: Oh, actually, we may need `epoch::pin()` in some of those places...

jonhoo:
Ah, no, you're right, with the 2018 edition you may be able to remove them. The one thing you _will_ need somewhere is probably this though
```rust
#[cfg(feature = "std")]
use std as alloc;
```
But could very well be that we can get rid of `cfg-if` entirely!

GarkGarcia:
> ```rust
> use std as alloc
> ```

The compiler still complains about `use std as alloc` being unused. Is this supposed to be placed at `lib.rs`?

jonhoo:
Hmm, I was thinking of elsewhere in the crate where we try to use `alloc::`, but apparently that's just now dealt with automatically? I guess just remove it then? What happens if you try to compile it locally now with `--no-default-features`?

GarkGarcia:
> 
> 
> Hmm, I was thinking of elsewhere in the crate where we try to use `alloc::`, but apparently that's just now dealt with automatically? I guess just remove it then? What happens if you try to compile it locally now with `--no-default-features`?

I get a bunch of errors related to the absence of some things auto-imported from `std`. Also, we're missing `epoch::pin` in a bunch of places.

I'll work on the auto-imported things. I don't see a clear path on how to deal with `epoch::pin` though. If we were to simply enforce `cfg(feature = "std")` the API would get a bit unusable in a `no_std` context...

jonhoo:
I _think_ anything where we _take_ a `&epoch::Guard` is fine in `no_std`, but anything where we have to construct a `epoch::Guard` ourselves is a no-go in that environment (sadly).

jonhoo:
I am very surprised that CI doesn't catch that it doesn't compile with `--no-default-features`. That _should_ certainly cause a CI error.

jonhoo:
It [looks](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=542&view=logs&j=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&t=c4c46c3f-f7ff-55f3-7964-b30fdf7483cd) like it doesn't even re-compile when `--no-default-features` is passed..

GarkGarcia:
> 
> 
> It [looks](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=542&view=logs&j=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&t=c4c46c3f-f7ff-55f3-7964-b30fdf7483cd) like it doesn't even re-compile when `--no-default-features` is passed..

Yep.

jonhoo:
If you do the following locally, in order, do you get any errors?
```console
$ cargo check --all --bins --examples
$ cargo check --all --bins --examples --no-default-features
```

GarkGarcia:
There is no clear alternative for `Vec` in `alloc`, which is used in `src/raw/mod.rs`. Should we go with a `LinkedList` then? Whatever we need, we need it to implement `From<Box<[_]>>`, so I think it should be contiguous.

GarkGarcia:
> 
> 
> If you do the following locally, in order, do you get any errors?
> 
> ```
> $ cargo check --all --bins --examples
> $ cargo check --all --bins --examples --no-default-features
> ```

Both run fine:

```
Finished dev [unoptimized + debuginfo] target(s) in 0.18s
```

GarkGarcia:
> 
> 
> There is no clear alternative for `Vec` in `alloc`, which is used in `src/raw/mod.rs`. Should we go with a `LinkedList` then? Whatever we need, we need it to implement `From<Box<[_]>>`, so I think it should be contiguous.

Maybe `VecDeque` is a better fit, since we're only the `Vec` for iteration I think.

GarkGarcia:
> 
> 
> I _think_ anything where we _take_ a `&epoch::Guard` is fine in `no_std`, but anything where we have to construct a `epoch::Guard` ourselves is a no-go in that environment (sadly).

I think we should then consider passing the guard as a parameter in some of those places. What do you think?

jonhoo:
> There is no clear alternative for `Vec` in `alloc`

[`alloc::vec::Vec`](https://doc.rust-lang.org/alloc/vec/struct.Vec.html)?

> Both run fine

Well, that is unexpected indeed.. What if you don't include the other `--` flags?

> I think we should then consider passing the guard as a parameter in some of those places. What do you think?

I think @cuviper is coming to the rescue with https://github.com/jonhoo/flurry/pull/45 !

GarkGarcia:
> 
> 
> > There is no clear alternative for `Vec` in `alloc`
> 
> [`alloc::vec::Vec`](https://doc.rust-lang.org/alloc/vec/struct.Vec.html)?
> 
> > Both run fine
> 
> Well, that is unexpected indeed.. What if you don't include the other `--` flags?
> 
> > I think we should then consider passing the guard as a parameter in some of those places. What do you think?
> 
> I think @cuviper is coming to the rescue with #45 !

Sorry hehe. I was looking a `alloc::collections`.

GarkGarcia:
Alright, we now have three problems:

* What should we do about the default value of `S`?

* How to handle the absence of `pin`? Make APIs std-only vs. pass the guard as a parameter.

* How to handle the stride without knowing the number of CPUs?

jonhoo:
> What should we do about the default value of `S`?

Not sure about this one. The "solution" in `hashbrown` is to use `ahash`, which can be used on `no_std` with [some caveats](https://github.com/rust-lang/hashbrown/issues/124)
https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L11-L17
If we went that route, we would also have to do https://github.com/rust-lang/hashbrown/pull/125. I don't love it. I _think_ one possibility is to have
```rust
#[cfg(std)]
pub type DefaultHashBuilder = std::collections::hash_map::RandomState;
#[cfg(not(std))]
pub enum DefaultHashBuilder {}
```
Maybe @Amanieu has some insight that could help here? It looks like @0ndorio may also be a good person to ping?

> How to handle the absence of `pin`? Make APIs std-only vs. pass the guard as a parameter.

See #45. I think any method that internally constructs a `epoch::pin` will be `std`-only, and that's fine with the proposed `HashMapRef`.

> How to handle the stride without knowing the number of CPUs?

Assume `NCPU = 1` I think.

jonhoo:
CI should now catch the `no_std` issues again btw!

GarkGarcia:
> CI should now catch the `no_std` issues again btw!

Great.

> See #45. I think any method that internally constructs a `epoch::pin` will be `std`-only, and that's fine with the proposed `HashMapRef`.

There are definitively places where we are not able to provide the same APIs in a `no_std` environment. However, I feel like there are places where parameterizing the guard wouldn't hurt, aka places where we're pinning to avoid adding an unnecessary parameter to the method (because pining wasn't an issue before we started working on this).

I don't know how effective that would be, because maybe pinning is actually necessary, or maybe there wouldn't be a way to construct the guard in that environment anyway. I just think we should avoid restricting the `no_std` APIs as much as possible.

I'll take a look at #45 tomorrow.

> Assume `NCPU = 1` I think.

Alright, done.

> Not sure about this one. The "solution" in `hashbrown` is to use `ahash`, which can be used on `no_std` with [some caveats](https://github.com/rust-lang/hashbrown/issues/124)
> https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L11-L17
> If we went that route, we would also have to do [rust-lang/hashbrown#125](https://github.com/rust-lang/hashbrown/pull/125). I don't love it. I _think_ one possibility is to have
> 
> ```rust
> #[cfg(std)]
> pub type DefaultHashBuilder = std::collections::hash_map::RandomState;
> #[cfg(not(std))]
> pub enum DefaultHashBuilder {}
> ```

I don't really fancy using an empty enum because I fell like we are abusing the type system, but it's a possibility. We would also need to implement `BuildHasher` for that empty enum, which feels silly. The `ahash` alternative seems cleaner to me.

Another possibility is to declare `HashMap` multiple times (one for `std` and one `no_std`) and only use the default parameter for `std`. Though that would force us to manually make sure the declarations match in future, as well as the associated documentation.

We may be better off simply omitting the default parameter.

jonhoo:
> However, I feel like there are places where parameterizing the guard wouldn't hurt, aka places where we're pinning to avoid adding an unnecessary parameter to the method (because pining wasn't an issue before we started working on this).

Yes, I agree with you that in those cases, we should change the method to instead take a `Guard`, and then rely on #45 as the way to access those methods _without_ a guard.

> We would also need to implement `BuildHasher` for that empty enum, which feels silly.

We actually don't, that's part of the trick. The idea here is that if you are not on `std`, then you _must_ specify `S` (since the default value does not produce a usable map), whereas if you are on `std`, then things "just work". It's actually kind of nice. I don't actually think there _is_ a reasonable default on `no_std`, and this is a way for us to have it that way without resorting to:

> We may be better off simply omitting the default parameter.

I _don't_ think that is a good idea. It is an _absolute_ pain if you are on `std` and you need to import and specify `RandomState` or `DefaultHashBuilder` _everywhere_.

> The `ahash` alternative seems cleaner to me.

I believe that still has the same issue, and would require some kind of fallback in case the user does _not_ enable the `ahash` feature.

> Another possibility is to declare `HashMap` multiple times (one for `std` and one `no_std`) and only use the default parameter for `std`.

I don't _think_ that will work, because cargo assumes that crate features are additive, and may compile a crate that did _not_ enable a feature _with_ that feature enabled. I'm having a hard time thinking of example code where _adding_ a default value to a crate that didn't expect it would break compilation, but I'm also not convinced it couldn't happen. We have to be _really_ careful with features.

cuviper:
> > Another possibility is to declare `HashMap` multiple times (one for `std` and one `no_std`) and only use the default parameter for `std`.
> 
> I don't _think_ that will work, because cargo assumes that crate features are additive, and may compile a crate that did _not_ enable a feature _with_ that feature enabled. I'm having a hard time thinking of example code where _adding_ a default value to a crate that didn't expect it would break compilation, but I'm also not convinced it couldn't happen. We have to be _really_ careful with features.

We did it like that in indexmap:
https://github.com/bluss/indexmap/blob/d3eb7cd5074ae26e2d77da433cad673c8a1a50e2/src/map.rs#L295-L306

That's automatically depending on `autocfg`-detected presence of `std`, but that detection won't work when you need to enable the feature in `crossbeam-epoch` too. (Build scripts can't change features.) I think it is still sufficiently additive for cargo features though. Adding a default parameter shouldn't hurt those `no_std` who were dealing with having no default.

jonhoo:
@cuviper That's good enough for me. In that case, I think we should do exactly like `indexmap` and not set a default in `no_std`.

Amanieu:
> https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L11-L17

That's just a hack for building hashbrown as part of libstd, in which case hashbrown cannot depend on any other crates. Since the `ahash` feature is enabled by default, even in no_std, we simply use ahash in all cases which works fine.

Amanieu:
My recommendation would be to pick a default hash function that works on no_std (e.g. fxhash or ahash) and let users opt-in to std's RandomState if they need HashDoS resistance.

GarkGarcia:
> Yes, I agree with you that in those cases, we should change the method to instead take a `Guard`, and then rely on #45 as the way to access those methods _without_ a guard.

Cool! I went ahead and worked on this. Take a look and see if there's any place where you feel like I should have made the method std-only instead of passing the guard as a parameter.

> I _don't_ think that is a good idea. It is an _absolute_ pain if you are on `std` and you need to import and specify `RandomState` or `DefaultHashBuilder` _everywhere_.

Yeah, you're right, that's pretty annoying.

> My recommendation would be to pick a default hash function that works on no_std (e.g. fxhash or ahash) and let users opt-in to std's RandomState if they need HashDoS resistance.

That makes more sense to me. Every other solution feels a bit hacky (though I wouldn't mind using if necessary to resolve this).

GarkGarcia:
Just to recap, if we consider the `epoch::pin` thing solved, our only problems are:

* How to deal with the default parameter of `S`?

* I'm not sure our tests are suited for `no_std`

Amanieu:
I wouldn't worry about testing under no_std except for just checking that it compiles. It's fine for tests to depend on std. For the compilation test you can just try cross-compiling to a thumb target in CI, which doesn't have std.

GarkGarcia:
_Side note:_ This default parameter issue we're facing feels like something somebody should write a macro for.

src/map.rs
```diff
@@ -172,16 +173,16 @@ where
         h.finish()
     }
 
+    #[inline]
     /// Tests if `key` is a key in this table.
     ///
     /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
     /// form must match those for the key type.
-    pub fn contains_key<Q>(&self, key: &Q) -> bool
+    pub fn contains_key<'g, Q>(&self, key: &Q, guard: &'g Guard) -> bool
```
jonhoo: This doesn't need `'g`, Just taking `&Guard` should be fine since we don't return a reference.

src/map.rs
```diff
@@ -261,19 +262,19 @@ where
...
 
+    #[inline]
     /// Obtains the value to which `key` is mapped and passes it through the closure `then`.
     ///
     /// Returns `None` if this map contains no mapping for `key`.
     ///
     /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
     /// form must match those for the key type.
-    pub fn get_and<Q, R, F>(&self, key: &Q, then: F) -> Option<R>
+    pub fn get_and<'g, Q, R, F>(&self, key: &Q, then: F, guard: &'g Guard) -> Option<R>
```
jonhoo: Same here -- no `'g` needed. Alternatively you could _also_ add the `'g` to `&self`, and then make `F: FnOnce(&'g V) -> R`.

src/map.rs
```diff
@@ -1069,11 +1070,9 @@ where
     #[inline]
     /// Tries to reserve capacity for at least additional more elements.
     /// The collection may reserve more space to avoid frequent reallocations.
-    pub fn reserve(&self, additional: usize) {
+    pub fn reserve<'g>(&self, additional: usize, guard: &'g Guard) {
```
jonhoo: No `'g` necessary.

src/map.rs
```diff
@@ -1265,16 +1264,15 @@ where
     /// If `f` returns `false` for a given key/value pair, but the value for that pair is concurrently
     /// modified before the removal takes place, the entry will not be removed.
     /// If you want the removal to happen even in the case of concurrent modification, use [`HashMap::retain_force`].
-    pub fn retain<F>(&mut self, mut f: F)
+    pub fn retain<'g, F>(&mut self, mut f: F, guard: &'g Guard)
```
jonhoo: No `'g` necessary. Or add to `&self` and then also include `'g` in the arguments to `FnMut`.

src/map.rs
```diff
@@ -1285,15 +1283,14 @@ where
     ///
     /// This method always deletes any key/value pair that `f` returns `false` for,
     /// even if if the value is updated concurrently. If you do not want that behavior, use [`HashMap::retain`].
-    pub fn retain_force<F>(&mut self, mut f: F)
+    pub fn retain_force<'g, F>(&mut self, mut f: F, guard: &'g Guard)
```
jonhoo: Same comment as for `retain`.

src/map.rs
```diff
@@ -1525,15 +1528,22 @@ where
 }
 
 #[inline]
+#[cfg(feature = "std")]
 /// Returns the number of physical CPUs in the machine (_O(1)_).
```
jonhoo: Why does this say `(_O(1)_)`?

GarkGarcia: My idea was to indicate that this takes constant time, and the only reason why this is not a `cons` is because `NCPU` need to initialized.

jonhoo:
> My recommendation would be to pick a default hash function that works on no_std (e.g. fxhash or ahash) and let users opt-in to std's RandomState if they need HashDoS resistance.

I'm torn on this. `IndexMap` and `std::collections::HashMap` default to `RandomState`, `hashbrown` defaults to `ahash`. I'm not sure which "default" category we fall into. It comes down to whether we want "DoS resistant but slower" by default or "non-DoS resistant and faster". Thinking ahead a little, I believe the "treeify" optimization from the Java code will actually make the code pretty DoS resistant automatically, so a non-DoS hasher by default actually makes a lot of sense. So, let's go with that. @Amanieu thanks for the input! Did you have a particularly good reason for going with `ahash` over `fxhash`?

> I wouldn't worry about testing under `no_std` except for just checking that it compiles. It's fine for tests to depend on `std`.

Yup, that's what CI currently does. Tests are run with all features enabled.

> For the compilation test you can just try cross-compiling to a thumb target in CI, which doesn't have std.

Ah, that's a good idea. This is one step beyond `cargo check --no-default-features`, and trying to compile on a target that doesn't even _have_ `std`. @GarkGarcia just add this to `azure-pipelines.yml`:
```yaml
 - job: no_std
   displayName: "Compile-check on no_std target"
   pool:
     vmImage: ubuntu-16.04
   steps:
     - template: install-rust.yml@templates
     - bash: cargo check --target thumbv7m-none-eabi --no-default-features
       displayName: cargo check
```

src/map.rs
jonhoo: Ah, this came across as "the number of physical CPUs is O(1)", which doesn't make much sense. I would probably just delete that part of the comment :)

azure-pipelines.yml
```diff
@@ -29,7 +29,14 @@ jobs:
...
        displayName: cargo deny
-
+ - job: no_std
+   displayName: "Compile-check on no_std target"
+   pool:
+     vmImage: ubuntu-16.04
+   steps:
+     - template: install-rust.yml@templates
+     - bash: cargo check --target thumbv7m-none-eabi --no-default-features
+       displayName: cargo check
```
jonhoo: You _might_ still need an empty line below here, I'm not sure.

GarkGarcia: Ok.

src/map.rs
GarkGarcia: No problem.

jonhoo:
CI probably isn't running because of the conflicts with master would be my guess. Either that or Azure is just super busy.

GarkGarcia:
> I'm torn on this. `IndexMap` and `std::collections::HashMap` default to `RandomState`, `hashbrown` defaults to `ahash`. I'm not sure which "default" category we fall into. It comes down to whether we want "DoS resistant but slower" by default or "non-DoS resistant and faster". Thinking ahead a little, I believe the "treeify" optimization from the Java code will actually make the code pretty DoS resistant automatically, so a non-DoS hasher by default actually makes a lot of sense. So, let's go with that. @Amanieu thanks for the input! Did you have a particularly good reason for going with `ahash` over `fxhash`?

Great! Whatever hasher we use I'll include a note in the docs about DoS attacks and cryptographical security. On regards to which one to use, `ahash` is apparently DoS resistant (as stated in the **Goals** section of the `README`), so maybe `fxhash` is a better fit?

jonhoo:
I think `ahash` is DoS resistant only with the randomness feature enabled. Given that we probably don't even need it though, we should probably depend on `ahash` with `default-features = false`.

Amanieu:
> Did you have a particularly good reason for going with ahash over fxhash?

I've found that FxHash can produce hash collisions on pretty common workloads (i.e. we got those by accident, not even with an attacker), to the point where hashtable performance is severely degraded. In terms of complexity, ahash is only a few more instructions, but avoids these collisions. It is also very well optimized for hashing strings.

> I think ahash is DoS resistant only with the randomness feature enabled.

I personally don't trust ahash to be DoS resistant, but it's fine for me since I don't require that for hashbrown.

GarkGarcia:
> 
> 
> I think `ahash` is DoS resistant only with the randomness feature enabled. Given that we probably don't even need it though, we should probably depend on `ahash` with `default-features = false`.

Ok then. Now our only problem is that `HashMap` don't implement `Debug` in a `no_std` environment (because we can't call `HashMap::iter`) and we've included `#![deny(missing_debug_implementations)]` in `lib.rs`.

jonhoo:
Yeah, that's awkward, but I think it's acceptable.

jonhoo:
As in, just do:
```rust
#![cfg_attr(feature = "std", deny(missing_debug_implementations))]
```
(and remove the existing `deny(missing_debug_implementations)`)

jonhoo:
Another alternative is to have a separate impl on `no_std` that just prints `HashMap`.

Amanieu:
Actually there is a potential source of unsoundness here: `crossbeam_epoch` allows you to create separate `Collector` instances aside from the global one used by `epoch::pin`. However you need to make sure that all `Guard`s used all come from the same `Collector`.

The approach I went with for crossbeam-skiplist is that the [`SkipList`](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-skiplist/src/base.rs) type is the low-level interface which must be associated with a `Collector` when created and must have a `Guard` passed in for every operation. The [`SkipMap`](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-skiplist/src/map.rs) type is the high level interface which hides all the epoch-related stuff from you, but cannot be used in no_std mode.

jonhoo:
@Amanieu oh, that is interesting. I believe that's a problem beyond just `no_std` though -- it applies to all the methods we provide that take a `&Guard`? @cuviper this suggests to me that we want to "invert" what you proposed in #44, and follow a design like the one Amanieu proposed above where the low level takes a `Collector`, and the high level manages the `Collector` for you.

Doesn't that mean that users cannot take advantage of the global epoch collection though? As in, they can't share the collection between many different programs that use `epoch::pin`.

Amanieu:
The high-level wrapper can use the default collector (the one used by `epoch::pin`):
```rust
    pub fn new() -> SkipMap<K, V> {
        SkipMap {
            inner: base::SkipList::new(epoch::default_collector().clone()),
        }
    }
```

jonhoo:
That does mean that methods cannot generally return references though, right? Since you would not be able to give the return value an appropriate lifetime, as it would be tied to the guard we construct _inside_ each method. That's pretty unfortunate.

jonhoo:
That is, in the high-level type. In the low-level type you can still take a `&Guard` and just assert that the collector matches (like it looks as though you're doing in `SkipList`)

cuviper:
A newtype `Guard` could be enforced to only be constructed the way you want. Not sure what that would be on `no_std` though.

Two of the motivations for #44 are `Index` and `IntoIterator for &HashMap`, which can't work with implicit temporary guards.

jonhoo:
Actually, I have a proposal. Writing it up as an issue as we speak. Will link when written.

jonhoo:
Okay, for the unsoundness let's continue that conversation in https://github.com/jonhoo/flurry/issues/46. I think this PR is mostly orthogonal to that issue.

@GarkGarcia for this PR, I think the main remaining thing is to merge `master`

GarkGarcia:
> @GarkGarcia for this PR, I think the main remaining thing is to merge `master`

Could you handle that for me please?

jonhoo:
Okay, so, I think the next (last?) challenge is that `parking_lot` does _not_ support `no_std`. We may have to be generic over the type of lock to use...

jonhoo:
All right, I fixed that by making us generic over `lock_api::RawMutex`. It ain't pretty, but it works. See also the commit message on https://github.com/jonhoo/flurry/pull/44/commits/adb58e03352f8dcef085faf97b33b4c7f34c8110 for details. This introduced another "default only on `std`" type parameter, where I opted to go for the `indexmap` approach of simply not setting a default without `std`.

jonhoo:
:tada: tests pass :tada: 

I'll merge this in a few hours unless there are any objections. The `L` parameter is unfortunate, but I don't see how we get around it if we want to support `no_std`. I _think_ the only question is "is it important is it for us to support `no_std` for *this* crate". If the answer is that it seems valuable, then I think this is how it has to be.

src/map.rs
```diff
@@ -60,14 +58,25 @@ macro_rules! load_factor {
...
 ///
+/// Note that `ahash::RandomState`, the default value of `S`, is not
+/// cryptographically secure. Therefore it is strongly recommended that you do
+/// not use this hash for cryptographic purproses.
+/// See [`ahash`](https://github.com/tkaitchuck/ahash) for more information.
+///
 /// See the [crate-level documentation](index.html) for details.
-pub struct HashMap<K, V, S = RandomState> {
+#[cfg(feature = "std")]
+pub struct HashMap<K, V, L = parking_lot::RawMutex, S = RandomState>
```
cuviper: I suggest `L` and `S` should be reversed, so `HashMap<K, V, S>` still looks like the standard collection. This would also simplify the test cases where you had to start specifying the lock type to use a different hasher, which feels like an API wart. However, that means you'd have to remove the default `S` in the no-std configuration too.

src/iter/iter.rs
```diff
@@ -1,17 +1,23 @@
...
 use crossbeam_epoch::Guard;
-use std::sync::atomic::Ordering;
 
 /// An iterator over a map's entries.
 ///
 /// See [`HashMap::iter`](crate::HashMap::iter) for details.
 #[derive(Debug)]
-pub struct Iter<'g, K, V> {
-    pub(crate) node_iter: NodeIter<'g, K, V>,
+pub struct Iter<'g, K, V, L>
```
cuviper: Maybe add a default `L` for all of the iterator types too?

src/map.rs
jonhoo: Yeah, not wanting to remove the default `S` in no-std is why I ordered them this way. I'm honestly not sure which is better/worse. My inclination is to make things harder for `no_std` folks when given a trade-off, which would suggest no defaults in no-std and `L` last.

src/iter/iter.rs
jonhoo: Bah, then we'd have to duplicate all of these too. One other option here is to define a default value on `no_std` that panics when used. It ain't great at all, but it would allow us to not duplicate all the structs.

jonhoo:
I don't disagree with you that the change is becoming pretty painful API-wise. I've put out a [call for input](https://twitter.com/jonhoo/status/1222942509606653952) from the people who _actually_ use `no_std` to see if we can make a more informed decision about whether the pain here is worth it.

If we decide it's not worth it, then I still think there are a bunch of changes from this PR we may want to adopt in `master` (like the explicit `Guard` passing in many places).

src/iter/iter.rs
cuviper: I suppose that's why `hashbrown` uses an uninhabited default on no-std.

Amanieu: No, hashbrown uses ahash on no_std. It only uses an uninhabited default when compiled as a dependency for libstd (because the libstd build system is weird and makes dependencies difficult).

cuviper: Oh ok, I misunderstood that context, and hashbrown doesn't need a lock like this. An uninhabited default might still work though, or we could use `macro_rules!` to hide the duplication.

jonhoo:
Let's continue this in https://github.com/jonhoo/flurry/pull/48.

PR #45, opened 2020-01-29
Add a struct HashMapRef which holds a reference and guard

cuviper:
This is created by `HashMap::as_ref(&self)`, and it implements most of
the same public API, just with implicit epoch guards. Additionally, it
implements `IntoIterator` by reference, and a new trick -- `Index`!


cuviper:
> it implements most of the same public API

Obvious absences are all of the `HashMap` constructors. In addition, I didn't add `retain` and `retain_force` because those take `&mut self` instead of `&self` -- was that intentional? They talk about concurrent access, but `&mut` prevents that entirely.

I also need to add tests, but I wanted to float the idea in a PR before I continue.

jonhoo:
Ah, that is just beautiful!

No, `&mut self` for `retain` was not intentional. Feel free to fix in this PR.

src/map_ref.rs
```diff
@@ -0,0 +1,214 @@
...
+use crate::HashMap;
+use crossbeam_epoch::{self as epoch, Guard};
+use std::borrow::Borrow;
+use std::collections::hash_map::RandomState;
+use std::fmt::{self, Debug, Formatter};
+use std::hash::{BuildHasher, Hash};
+use std::ops::Index;
+
+/// A reference to a [`HashMap`], constructed with [`HashMap::as_ref`].
+/// The current thread will be pinned for the duration of this reference.
```
jonhoo: I think it is worthwhile to document the implications of that here, since they are pretty severe. Specifically, no destructors will be run for as long as this lives!

src/map_ref.rs
```diff
@@ -0,0 +1,214 @@
...
+pub struct HashMapRef<'map, K, V, S = RandomState> {
+    map: &'map HashMap<K, V, S>,
+    guard: Guard,
+}
+
+impl<K, V, S> HashMap<K, V, S> {
+    /// Get a reference to this map with the current thread pinned.
```
jonhoo: Let's also point people to the type-level docs for `HashMapRef` here.

src/map_ref.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub fn as_ref(&self) -> HashMapRef<'_, K, V, S> {
```
jonhoo: I wonder -- do we _also_ want a constructor that _takes a `&epoch::Guard`? That'd be handy for #44

jonhoo:
I pushed a fix for `retain` taking `&mut` to `master`.

Also, https://github.com/jonhoo/flurry/pull/44 just added explicit guards to a bunch more methods, which may be a good precursor to the work here.

tests/basic_ref.rs
```diff
@@ -0,0 +1,362 @@
...
+        assert!(kv == (&i, &0) || kv == (&i, &1));
+    }
+}
+
+#[test]
+fn concurrent_remove() {
+    let map = Arc::new(HashMap::<usize, usize>::new());
+
+    {
+        let map = (*map).as_ref();
```
cuviper: I'm not thrilled with this myself -- I think it's because `<Arc as AsRef>::as_ref` takes precedence over the deref to our inherent `HashMap::as_ref`. Maybe we need a different name, like `HashMap::guarded()`. A name like that might pair better with a user's guard too, `HashMap::with_guard(guard: &guard)`.

cuviper:
Hmm, maybe I should just cross-link docs, instead of duplicating them...

jonhoo:
Yeah, I think that's probably fine, given that they are "feed-through" anyway. I wish there was a way to `#[doc(include(path::to::type::or::method))]`.

jonhoo:
Oh, also, you may want to rebase on/merge `master` given that I merged #47 which included a bunch more guarded methods.

src/map_ref.rs
cuviper: Can't they just click the link on the return value?

src/map_ref.rs
```diff
@@ -0,0 +1,272 @@
...
+where
+    K: Sync + Send + Clone + Hash + Eq,
+    V: Sync + Send,
+    S: BuildHasher,
+{
+    /// Get a reference to this map with the current thread pinned.
+    ///
+    /// Keep in mind that for as long as you hold onto this, you are preventing the collection of
+    /// garbage generated by the map.
+    pub fn guarded(&self) -> HashMapRef<'_, K, V, S> {
```
jonhoo: I wonder if we want to call this method `pin` to indicate that that's _really_ what you're doing.

src/map_ref.rs
```diff
@@ -0,0 +1,272 @@
...
+use std::hash::{BuildHasher, Hash};
+use std::ops::{Deref, Index};
+
+/// A reference to a [`HashMap`], constructed with [`HashMap::guarded`] or [`HashMap::with_guard`].
+///
+/// The current thread will be pinned for the duration of this reference.
+/// Keep in mind that this prevents the collection of garbage generated by the map.
+pub struct HashMapRef<'map, K: 'static, V: 'static, S = crate::DefaultHashBuilder> {
+    map: &'map HashMap<K, V, S>,
+    guard: GuardRef<'map>,
```
jonhoo: I think we may also want to have
```rust
pub fn repin(&mut self)
```
Which calls [`Guard::repin`](https://docs.rs/crossbeam-epoch/0.8.0/crossbeam_epoch/struct.Guard.html#method.repin).

src/map_ref.rs
cuviper: I was actually a little surprised on `fn guard(&self) -> Guard` that you didn't call that `pin`, since it ends with calling `pin()` inside. But I'm ok if you want to call this `pin` instead.

src/map_ref.rs
cuviper: We can do that in the `Owned` case, but we don't have exclusive access to the guard in the `Ref` case from `with_guard`. Would you have `repin` be a no-op in that case? Or should we return a `bool` whether we were able to actually call `repin`?

Or we could just expose the guard with a couple of methods, so you can `repin` it yourself.

```rust
    fn guard(&self) -> &Guard;
    fn guard_mut(&mut self) -> Option<&mut Guard>;
```

e.g. `map_ref.guard_mut().map(Guard::repin)`

src/map_ref.rs
jonhoo: So I actually didn't call it `pin` specifically because I wanted _this_ method to be called `pin`. My thinking here is that it'll often make _more_ sense to `pin` and then operate on a `HashMapRef` than to manually get a `Guard` and then pass that to a bunch of methods. Making the `pin` method return a `HashMapRef` makes it more likely that people will reach for and discover this way of using the map, which I think we want to encourage.

src/map_ref.rs
jonhoo: Ah, damn, you're right...Hmm. Maybe the `repin` use case is specific enough that if you want that behavior you have to use the non-`HashMapRef` interface. That seems fine. The methods you gave above would work, but I think in practice they'd probably go essentially unused.

jonhoo:
Fantastic, thank you!

PR #47, opened 2020-01-30
Steal good commits from #44

jonhoo:
This steals a bunch of the commits from #44 that are _not_ related to `no_std`, and should land independently of whether we merge `no_std` support.


GarkGarcia:
Awesome! I'm happy I could help 😃

PR #48, opened 2020-01-30
Make the crate `no_std` + `alloc` (rebased)

### LINKED ISSUE #26: no_std + alloc

jonhoo:
This is effectively a squash + rebase of #44 on top of #47.

It squashes the commits so that it is easier to see what's going on. @cuviper's comments from https://github.com/jonhoo/flurry/pull/44#pullrequestreview-351018649 still apply, as does the question about whether we are even willing to make this change (https://github.com/jonhoo/flurry/pull/44#issuecomment-580378812). There's a little twitter straw poll [here](https://twitter.com/jonhoo/status/1222942509606653952). To summarize:

> The new `L` parameter is unfortunate, but I don't see how we get around it if we want to support `no_std`. I think the only question is "is it important is it for us to support `no_std` for this crate". If the answer is that it seems valuable, then I think this is how it has to be.

Thanks to @GarkGarcia for all the hard work on #44 -- that's the majority of this change.

Closes #44. Fixes #26.


jonhoo:
Pushed some changes now that I think makes this patch a bit more pleasant. There is still the `L` parameter, but it should be much less noticeable since most users should be interacting with the `HashMap` type alias, which does not have it.

jonhoo:
Not sure if this alleviates some of your concerns from #44 @cuviper?

GarkGarcia:
> The new `L` parameter is unfortunate, but I don't see how we get around it if we want to support `no_std`. I think the only question is "is it important is it for us to support `no_std` for this crate". If the answer is that it seems valuable, then I think this is how it has to be.

Can't we also provide a default to the `L` parameter? That would alleviate some of the pain. No idea what that default would be in `no_std` though.

jonhoo:
Yeah, that's the problem -- on `no_std`, that parameter has no reasonable default. That in turn means that it cannot be the last type parameter (since you cannot have a type parameter without a default _after_ one with a default). Which again in turn means that anywhere people specify map types, if they want to use a custom hasher, they have to write `HashMap<K, V, _, S>`. Also, the extra generic type is real awkward for anyone who wants to be generic over a map, though that's something we won't completely get away from anyway.

cuviper:
I guess the type alias helps hide the complexity, at the loss of available iterator types. I'm not sure if that's a net win -- I guess that's still driven by how desirable `no_std` is in the first place.

Did you manage to gather compelling use-cases? And can you prototype that this crate actually works in those cases? What `L` would you use?

jonhoo:
I agree, it's definitely still a cost.

As far as use-cases goes, the response was pretty meager. https://twitter.com/thequux/status/1223248563523723264 and https://twitter.com/thequux/status/1223248563523723264 were the primary ones.

My current take is that we should probably just close this, with a note that we'd be open to considering if someone has a compelling use-case they're willing to also commit some time to supporting.

cuviper:
> My current take is that we should probably just close this, with a note that we'd be open to considering if someone has a compelling use-case they're willing to also commit some time to supporting.

I think that makes sense.

jonhoo:
All right, I'll close this and the corresponding issue then :)

Issue #46, opened 2020-01-30
It is unsound to accept arbitrary user-constructed Guards


jonhoo:
As @Amanieu points out in https://github.com/jonhoo/flurry/pull/44#issuecomment-580290129, the current API which lets users supply their own `Guard` references to our methods is unsound. Specifically, a user can do the following:
```rust
use crossbeam_epoch::*;
let map = HashMap::default();
map.insert(42, String::from("hello"), &epoch::pin());

let evil = crossbeam_epoch::Collector::new();
let evil = evil.register();
let oops = map.get(&42, &evil.guard());

map.remove(&42, &epoch::pin());
// at this point, the default collector is allowed to free `"hello"`
// since no-one has the global epoch pinned as far as it is aware.
// `oops` is tied to the lifetime of a Guard that is not a part of
// the same epoch group, and so can now be dangling.
// but we can still access it!
assert_eq!(oops, "hello");
```

Here is a sketch of a possible solution, largely inspired by @Amanieu's proposal + #45.

First, we add a `Collector` field to `HashMap`, and add a constructor that takes a `Collector` (this can be `no_std`). The `default()` and `new()` constructors use `epoch::default_collector().clone()` as their `Collector`. Then, we add a check to every method that takes a `&Guard` (which is essentially all of them at the moment) that checks that the given `Guard`'s collector matches `self.collector` (see [`SkipList::check_guard`](https://github.com/crossbeam-rs/crossbeam/blob/ebecb82c740a1b3d9d10f235387848f7e3fa9c68/crossbeam-skiplist/src/base.rs#L371-L375)).  This should take care of the safety issues.

We then add a wrapper type that is tied to a _single_ `Guard` and exposes all the methods (and some more convenience ones like `Index`) without a `Guard` argument. This is #45. It should also have two constructors. One that takes a user-supplied `Guard` (and checks it), and one that does not (and uses `epoch::pin`). There should be a method to extract the `Guard` again from the `HashMapRef` (since it needs to take ownership of it to also have a variant that uses `epoch::pin`).

cuviper:
If we use a `Cow`-like enum, we can still just borrow a user's guard. (Not `std::borrow::Cow` directly since `Guard` can't "copy" with `ToOwned`.)

PR #49, opened 2020-01-30
Fix two unsoundness bugs

jonhoo:
Fixes both #46 and the issue highlighted [here](https://github.com/crossbeam-rs/crossbeam/blob/ebecb82c740a1b3d9d10f235387848f7e3fa9c68/crossbeam-skiplist/src/base.rs#L308-L319) (`K` and `V` must both be `'static` in case of a global collector!).

@cuviper I wonder if we could do something cool with this change now that we have the `Collector`. I'm specifically thinking something like `HashMap::pin`, which returns a `HashMapRef` (like in #45), which internally holds a `LocalHandle` and a `Guard` (not sure how to make the lifetimes work out).


cuviper:
Ugh, that's awkward that `Collector`s don't have an actual lifetime in them. I'm trying to think of other ways we can deal with this, but I'm not sure.

> I'm specifically thinking something like `HashMap::pin`, which returns a `HashMapRef` (like in #45), which internally holds a `LocalHandle` and a `Guard` (not sure how to make the lifetimes work out).

There are no lifetimes between those either, so it should work just fine to hold both in the same struct. I guess you don't even really need the `LocalHandle` at that point, just the `Guard`, unless you want to expose that handle to the user.

cuviper:
As a first, minimal step, I'd suggest just requiring `'static` types and assert that `Guard`s are from the global collector. Custom collectors complicate the correct usage of the map, as now you have to go through the long `register` path to get a guard. (Unless you _know_ it still has the default collector, but this isn't represented at all at the type level.)

jonhoo:
I think this change is strictly more general than it was previously? If you just use `HashMap::new` and friends (basically, if you never call `with_constructor`), then you can always just use `epoch::pin` like you did before (and not go through `register`). Or am I missing something?

It might be nice to provide `HashMap::guard` which just directly gives you a `Guard` too.

cuviper:
> If you just use `HashMap::new` and friends (basically, if you never call `with_constructor`), then you can always just use `epoch::pin` like you did before (and not go through `register`). Or am I missing something?

You can only assume that if you **know** that the map was never in the hands of anything that might have called `with_constructor`. That means both controlling `self` and `&mut self` (because of `swap`) at all times. Maybe that's paranoid of me, but I don't like that this is a transparent footgun.

jonhoo:
Yes, I think any code that _takes_ a `HashMap` must use `register`. You're not wrong that it's a potential foot-gun! How about we remove `with_collector` for the time being, given that (I think) that is the only thing currently blocking this? We still leave the `Collector` in there for the purposes of `HashMapRef`.

PR #54, opened 2020-01-31
Add HashMapRef::clear

cuviper:


jonhoo:
I wonder if there's a way for us to write some kind of simple check that `HashMapRef` provides the same public methods (by name) as `HashMap`... :thinking:

PR #55, opened 2020-02-02
Use `clear()` in benchmarks

domenicquirl:
The benchmarks adapted from `hashbrown` introduced by #43 were missing the `clear()` method on the map and thus currently construct a new map every iteration. Now that we have `clear()` thanks to #42, we should fix this and make the benchmarks identical.

### DIFF
--- a/benches/flurry_hashbrown.rs
+++ b/benches/flurry_hashbrown.rs
@@ -51,14 +51,14 @@ macro_rules! bench_suite {
 macro_rules! bench_insert {
     ($group:ident, $keydist:expr, $bench_id: expr) => {
         $group.bench_function(BenchmarkId::from_parameter($bench_id), |b| {
+            let map: HashMap<_, _> = HashMap::with_capacity(SIZE as usize);
             b.iter(|| {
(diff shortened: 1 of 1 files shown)


benches/flurry_hashbrown.rs
```diff
@@ -51,10 +51,10 @@ macro_rules! bench_suite {
 macro_rules! bench_insert {
     ($group:ident, $keydist:expr, $bench_id: expr) => {
         $group.bench_function(BenchmarkId::from_parameter($bench_id), |b| {
+            let mut map: HashMap<_, _> = HashMap::with_capacity(SIZE as usize);
```
jonhoo: I don't think this has to be `mut`?

jonhoo:
That's a lot of commits for a small change! Could you squash them?

domenicquirl:
Curious, for some reason this has the commits from adding the benchmarks in the first place again. I'll see if I can get rid of those...

jonhoo:
@domenicquirl I'm happy to merge this once the unnecessary `mut` is removed :)

domenicquirl:
I missed that comment entirely 😅 
Updated the commit to not include `mut` there.

Issue #50, opened 2020-01-30
Profile and improve map performance


domenicquirl:
We just merged #43, which gives us benchmarks! Currently we have benches against [`dashmap`](https://github.com/xacrimon/dashmap/tree/master/benches) and [`hashbrown`](https://github.com/rust-lang/hashbrown/tree/master/benches). 

First runs of the benchmarks seem to indicate that retrieving values from the map is quite performant, but inserting into the map is rather slow compared to both other maps. Preliminary results:

- **against hashbrown**, `insert` and `insert_erase` run in the order of on ms, compared to 10-20 us in `hashbrown`. `get` holds up reasonably well though, still ~20 us to ~5-6, but given the added concurrency this seems much better than `insert`. Both of these use one guard across all operations of one benchmark run and relatively consistent across input distributions.
- **against dashmap**, `dashmap` inserts in something between 1 ms with 1 thread and .8 ms for multiple threads, however what's probably more important is throughput. This for some reason is not in the report produced by `criterion`, but was on the order of 100 MElements/s. We compare with about 25-12 ms depending on thread count, which is so slow it takes several minutes to run any benchmark and puts throughput at something like 5, maybe 10 MElements/s. This is using 1 guard per map operation, which admittedly is not what you would do in a scenario such as the benchmark.
- **against dashmap, but with only one guard per thread** (the different setup), we `insert` in 14-11 ms, which is better, but still considerably slower than `dashmap`. Note that this does not compare directly due to the different setup of the threads.
- **get against dashmap** seems really good! `dashmap` gets 700-500 us, while we get 1.5 ms on one thread, 800us on 2 threads and go down to below 300 us on 8 threads, even with `pin()` for every call to `get`. With only one guard, this improves to 800 to a bit more than 200 us, depending on thread count.

It would be interesting to find out if these results are reproducible and, if so, what causes inserts to be so slow. @jonhoo suggests graphing the results of `perf record` as a profiling method [in the discussion in #43](https://github.com/jonhoo/flurry/pull/43#issuecomment-580464809), maybe this can be used as a starting point for the analysis.

jonhoo:
To start people off, here is a flamegraph of `insert_flurry_hashbrown/low_guard_once`:
[![flamegraph](https://jon.tsp.io/share/flurry-hashbrown-insert-low-guard-once.svg)](https://jon.tsp.io/share/flurry-hashbrown-insert-low-guard-once.svg)

[...]

PR #56, opened 2020-02-02
Share BinEntry::Moved across all moved bins in a table

domenicquirl:
This originates from #50 and tries to achieve performance improvements by instantiating only one `BinEntry::Moved` per table, as for all moved bins inside a table the `Moved` points to the same next table. In particular, this tries to reduce the amount of allocations of `Moved` by `Owned::new`, only `load`ing more shared pointers to the `Moved` instead.

This is submitted as a draft for the following reasons:

- ~~the history for this contains the changes in #55, which is still pending approval~~ (Edit: changed that)
- the previous implementation of `Table::drop` [contains a section](https://github.com/jonhoo/flurry/blob/b6b360feffa3ae6515739f6bf341fa27a7d0dc87/src/raw/mod.rs#L86-L98) which iterates through all bins, drops `Moved` entries and checks that none of the bins are non-empty. With these changes, the `Moved` is only dropped once (and turning all shared pointers to it `into_owned` would be incorrect), so this loop is not technically necessary any longer. I left the check against `Moved` in for now, but would like to discuss the possibility of removing it and relying on the other parts of the map implementation to uphold the invariant of all bins being empty at this point.
- I wanted to check against the results of sanitizer CI tests.
- in general, I want to give room for discussion on this.

I will add performance results to #50 as soon as I am done submitting this.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -328,7 +328,7 @@ where
         // swap happened, it must have happened _after_ we read. since we did the read while
         // pinning the epoch, the drop must happen in the _next_ epoch (i.e., the one that we
         // are holding up by holding on to our guard).
-        let node = unsafe { bin.deref() }.find(h, key, guard);
+        let node = table.find(unsafe { bin.deref() }, h, key, guard);
         if node.is_null() {
             return None;
         }
@@ -472,8 +472,8 @@ where
             // Safety: node is a valid pointer because we checked
             // it in the above if stmt.
             match unsafe { raw_node.deref() } {
-                BinEntry::Moved(next_table) => {
-                    table = self.help_transfer(table, *next_table, guard);
+                BinEntry::Moved => {
+                    table = self.help_transfer(table, guard);
                     // start from the first bin again in the new table
                     idx = 0;
                 }
@@ -574,7 +574,7 @@ where
             //  3. if table is set by a Moved node (below) through help_transfer, it will _either_
             //     keep using `table` (which is fine by 1. and 2.), or use the `next_table` raw
             //     pointer from inside the Moved. to see that if a Moved(t) is _read_, then t must
-            //     still be valid, see the safety comment on BinEntry::Moved.
+            //     still be valid, see the safety comment on Table.next_table.
             let t = unsafe { table.deref() };
 
             let bini = t.bini(h);
@@ -612,8 +612,8 @@ where
             // are holding up by holding on to our guard).
             let key = &node.as_node().unwrap().key;
             match *unsafe { bin.deref() } {
-                BinEntry::Moved(next_table) => {
-                    table = self.help_transfer(table, next_table, guard);
+                BinEntry::Moved => {
+                    table = self.help_transfer(table, guard);
                 }
                 BinEntry::Node(ref head)
                     if no_replacement && head.hash == h && &head.key == key =>
@@ -781,7 +781,7 @@ where
             //  3. if table is set by a Moved node (below) through help_transfer, it will _either_
             //     keep using `table` (which is fine by 1. and 2.), or use the `next_table` raw
             //     pointer from inside the Moved. to see that if a Moved(t) is _read_, then t must
-            //     still be valid, see the safety comment on BinEntry::Moved.
+            //     still be valid, see the safety comment on Table.next_table.
             let t = unsafe { table.deref() };
 
             let bini = t.bini(h);
@@ -807,8 +807,8 @@ where
             // pinning the epoch, the drop must happen in the _next_ epoch (i.e., the one that we
             // are holding up by holding on to our guard).
             match *unsafe { bin.deref() } {
-                BinEntry::Moved(next_table) => {
-                    table = self.help_transfer(table, next_table, guard);
+                BinEntry::Moved => {
+                    table = self.help_transfer(table, guard);
                 }
                 BinEntry::Node(ref head) => {
                     // bin is non-empty, need to link into it, so we must take the lock
@@ -947,19 +947,22 @@ where
     fn help_transfer<'g>(
         &'g self,
         table: Shared<'g, Table<K, V>>,
-        next_table: *const Table<K, V>,
         guard: &'g Guard,
     ) -> Shared<'g, Table<K, V>> {
-        if table.is_null() || next_table.is_null() {
+        if table.is_null() {
             return table;
         }
-
-        let next_table = Shared::from(next_table);
 
         // safety: table is only dropped on the next epoch change after it is swapped to null.
         // we read it as not null, so it must not be dropped until a subsequent epoch. since we
         // held `guard` at the time, we know that the current epoch continues to persist, and that
         // our reference is therefore valid.
+        let next_table = unsafe { table.deref() }.next_table(guard);
+        if next_table.is_null() {
+            return table;
+        }
+
+        // safety: same as above
         let rs = Self::resize_stamp(unsafe { table.deref() }.len()) << RESIZE_STAMP_SHIFT;
 
         while next_table == self.next_table.load(Ordering::SeqCst, guard)
@@ -1189,20 +1192,16 @@ where
             // are no longer active. these references are still active (marked by the guard), so
             // the target of these references won't be dropped while the guard remains active.
             let table = unsafe { table.deref() };
-            let next_table = unsafe { next_table.deref() };
 
             let bin = table.bin(i as usize, guard);
             if bin.is_null() {
                 advance = table
-                    .cas_bin(
-                        i,
-                        Shared::null(),
-                        table.get_moved(next_table as *const _, guard),
-                        guard,
-                    )
+                    .cas_bin(i, Shared::null(), table.get_moved(next_table, guard), guard)
                     .is_ok();
                 continue;
             }
+            // safety: as for table above
+            let next_table = unsafe { next_table.deref() };
 
             // safety: bin is a valid pointer.
             //
@@ -1219,7 +1218,7 @@ where
             // pinning the epoch, the drop must happen in the _next_ epoch (i.e., the one that we
             // are holding up by holding on to our guard).
             match *unsafe { bin.deref() } {
-                BinEntry::Moved(_) => {
+                BinEntry::Moved => {
                     // already processed
                     advance = true;
                 }
@@ -1310,7 +1309,10 @@ where
 
                     next_table.store_bin(i, low_bin);
                     next_table.store_bin(i + n, high_bin);
-                    table.store_bin(i, table.get_moved(next_table as *const _, guard));
+                    table.store_bin(
+                        i,
+                        table.get_moved(Shared::from(next_table as *const _), guard),
+                    );
 
                     // everything up to last_run in the _old_ bin linked list is now garbage.
                     // those nodes have all been re-allocated in the new bin linked list.
@@ -1521,7 +1523,7 @@ where
             //
             //  2. if table is set by a Moved node (below) through help_transfer, it will use the
             //     `next_table` raw pointer from inside the Moved. to see that if a Moved(t) is
-            //     _read_, then t must still be valid, see the safety comment on BinEntry::Moved.
+            //     _read_, then t must still be valid, see the safety comment on Table.next_table.
             let t = unsafe { table.deref() };
             let n = t.len() as u64;
             if n == 0 {
@@ -1548,8 +1550,8 @@ where
             // pinning the epoch, the drop must happen in the _next_ epoch (i.e., the one that we
             // are holding up by holding on to our guard).
             match *unsafe { bin.deref() } {
-                BinEntry::Moved(next_table) => {
-                    table = self.help_transfer(table, next_table, guard);
+                BinEntry::Moved => {
+                    table = self.help_transfer(table, guard);
                 }
                 BinEntry::Node(ref head) => {
                     let head_lock = head.lock.lock();
--- a/src/map.rs
+++ b/src/map.rs
@@ -1197,7 +1197,7 @@ where
                     .cas_bin(
                         i,
                         Shared::null(),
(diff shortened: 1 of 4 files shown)


src/raw/mod.rs
```diff
@@ -29,6 +35,23 @@ impl<K, V> Table<K, V> {
...
+        for_table: *const Table<K, V>,
+        guard: &'g Guard,
+    ) -> Shared<'g, BinEntry<K, V>> {
+        match self.moved.load(Ordering::SeqCst, guard) {
+            s if s.is_null() => {
+                // if a BinEntry::Moved associated with this table does not yet exist,
+                // create one and store it in `self.moved`
+                self.moved
+                    .store(Owned::new(BinEntry::Moved(for_table)), Ordering::SeqCst);
+                self.moved.load(Ordering::SeqCst, guard)
```
jonhoo: We can probably avoid the extra `load` here. Also, I wonder if the `store` should be a `compare_and_swap` in case this can race?

src/raw/mod.rs
```diff
@@ -29,6 +35,23 @@ impl<K, V> Table<K, V> {
...
+            }
+            s => s,
```
jonhoo: We should probably `assert!` here that `for_table` matches what is in `s`.

domenicquirl: Good point!

src/raw/mod.rs
domenicquirl: How would you get from `Owned` to `Shared` without this? Both `store` and `compare_and_swap` consume the stored pointer. I'm probably missing something, but I don't see it atm

jonhoo: [`Owned::into_shared`](https://docs.rs/crossbeam-epoch/0.8.0/crossbeam_epoch/struct.Owned.html#method.into_shared)

jonhoo:
Here's a question: do we even need to store the address inside of `Moved` anymore? Could `Table` just have a `next_table: Atomic<Table>` instead? The fact that the Java version takes care to store the pointer directly inside the `Moved` makes me think that maybe different `Moved` entries can have _different_ `next_table` pointers? Imagine that while walking a given table, _two_ resizes happen, could some of the `Moved` of the first table point directly to the newest table? Or do they have to point to intermediates (like this PR implicitly assumes) for stuff like iteration to work correctly?

domenicquirl:
Ah, the sanitizers fail due to a test case which manually constructs `Moved` and puts it into bins, of course they don't get freed then.

domenicquirl:
Assuming the `Moved` can be shared, we would only need `Moved` as a marker and _could_ then store a pointer to the next table directly (whether as `Atomic` or `*const`). However part of the idea of this change was to only instantiate `Moved` once per table, and if you store the address outside of `Moved`, this would create multiple instances of the marker again. I don't think we can get away without such a marker for moved bins.

With regard to different `next_table`s, I don't see how this would happen. The table stored in `Moved` is always `self.next_table`, which is set only at the start of a resize and reset to `null` when the resize finishes. Then the `Moved` are inserted into the current `self.table` (the old one, during the resize), and `self.table` and `self.next_table` does not change during such a resize. Then `self.table` _becomes_ `self.next_table`, so for a second resize, `Moved` will be inserted into the new table.

jonhoo:
Oh, you could still _also_ share the `Moved`, it just wouldn't contain any fields. If we can get rid of that `*const Table` and make `Moved` a true marker variant, that'd make me very happy.

Right, I agree with you that that makes sense, but the Java version very deliberately chooses to store the pointer in _every_ `Moved` (at least as far as I remember), and that gives me pause.

domenicquirl:
I think I found out why they did it. The reason I think is that `find` is implemented on `BinEntry` to handle the different kinds of nodes. Hence the node requires information about where to look in the `Moved` case to be contained within it. The alternative of course is to let `Table` do `find`.

jonhoo:
Ah, I see, interesting.. Or to pass `&Table` to `find`.

domenicquirl:
Okay so I did that (spending half the time rewriting test cases that got broken by all the moving around 😅 ). I stuck with `Table::find(bin)`, as `BinEntry::find(table)` just seems wrong to me. If information from `Table` is required, then `Table` should be the responsible type I feel, the other way is kind of "backwards".

Anyways, since the tests causing the asan fail were among the ones I touched anyways, here's hoping I didn't miss anything and the sanitizers stop complaining :P

domenicquirl:
Nice, fixed the tests. Please review again. Also, if we stick with some form of this change, the question about the checks in `Table::drop` is still open.

jonhoo:
Oh, yeah, I didn't expect that removing the pointer would change performance really, it's just one fewer "weird" thing in the code :)

src/node.rs
```diff
@@ -1,46 +1,14 @@
...
-    //     the call to `defer_destroy`. Thus, t remains valid for at least the lifetime of g.
-    //
-    //   - After releasing g, either the current resize is finished and operations on the map
-    //     cannot access t anymore as a more recent table will be loaded as the current table
-    //     (see once again `map::HashMap::transfer`), or the argument is as above.
-    //
-    // Since finishing a resize is the only time a table is `defer_destroy`ed, the above covers
-    // all cases.
-    Moved(*const Table<K, V>),
+    Moved(),
```
jonhoo: This should be just a marker variant now:
```suggestion
    Moved,
```

src/map.rs
```diff
@@ -947,15 +952,13 @@ where
     fn help_transfer<'g>(
         &'g self,
         table: Shared<'g, Table<K, V>>,
-        next_table: *const Table<K, V>,
+        next_table: Shared<'g, Table<K, V>>,
```
jonhoo: Does this even need to take a `next_table` argument now, or can it just load it from the passed `table`?

jonhoo: I'm _pretty_ sure that by the time this is called, `table.next_table` has already been set, so `t.next_table()` can be used over `t.get_moved()`.

src/map.rs
```diff
@@ -1310,7 +1309,10 @@ where
 
                     next_table.store_bin(i, low_bin);
                     next_table.store_bin(i + n, high_bin);
-                    table.store_bin(i, Owned::new(BinEntry::Moved(next_table as *const _)));
+                    table.store_bin(
+                        i,
+                        table.get_moved(Shared::from(next_table as *const _), guard),
```
jonhoo: Isn't `next_table` here _already_ a `Shared`?

src/raw/mod.rs
```diff
@@ -1,17 +1,59 @@
...
 pub(crate) struct Table<K, V> {
     bins: Box<[Atomic<BinEntry<K, V>>]>,
+
+    // since a Moved does not contain associated information,
+    // one instance is sufficient and shared across all bins in this table
+    moved: Atomic<BinEntry<K, V>>,
+    // since the information content of moved nodes is the same across
+    // the table, we share this information
+    //
+    // safety: next_table is a valid pointer if it was read as consequence of reading
```
jonhoo: I think this argument has to refer to the fact that the user has a reference to _this_ table.

src/map.rs
domenicquirl: This is in fact always what happens before and is passed as the argument. If you prefer, this could be handled by `help_transfer`

src/raw/mod.rs
```diff
@@ -76,26 +216,42 @@ impl<K, V> Table<K, V> {
...
         // lifetime of their borrow of self, so there cannot be any outstanding references to
         // anything in the map.
         let guard = unsafe { crossbeam_epoch::unprotected() };
 
+        // since BinEntry::Nodes are either dropped by drop_bins or transferred to a new table,
+        // all bins are empty or contain a Shared pointing to shared the BinEntry::Moved (if
+        // self.bins was not replaced by drop_bins anyway)
         for bin in &self.bins[..] {
             let bin = bin.swap(Shared::null(), Ordering::SeqCst, guard);
+            // TODO: do we need the safety check against moved or do we assume the above invariant here?
```
jonhoo: What safety check are you thinking of?

src/map.rs
domenicquirl: Here `next_table` is `deref`ed before. I moved this after the `if` in 1197 to avoid casting, but the immediately preceding operations take a reference. So either this or turn everything else into `unsafe { next_table.deref() }` inline I think

src/raw/mod.rs
```diff
@@ -76,26 +216,42 @@ impl<K, V> Table<K, V> {
...
         // safety: we have &mut self _and_ all references we have returned are bound to the
         // lifetime of their borrow of self, so there cannot be any outstanding references to
         // anything in the map.
         let guard = unsafe { crossbeam_epoch::unprotected() };
 
+        // since BinEntry::Nodes are either dropped by drop_bins or transferred to a new table,
+        // all bins are empty or contain a Shared pointing to shared the BinEntry::Moved (if
+        // self.bins was not replaced by drop_bins anyway)
         for bin in &self.bins[..] {
             let bin = bin.swap(Shared::null(), Ordering::SeqCst, guard);
```
jonhoo: A separate observation -- this loop is probably quite expensive due to the write here when all we're doing it checking that there aren't any `Node`s in it, and dropping any references. We could probably do better by just swapping out `bins` with an empty `Box<[...]>` instead.

src/raw/mod.rs
domenicquirl: Correct, in particular it should refer to _this_ table being loaded _as_ `HashMap.table`

src/map.rs
jonhoo: Previously I don't think `help_transfer` had any way to learn the address of the next table. Now it does, so I think we should take advantage of that!

src/raw/mod.rs
domenicquirl: That's what I mentioned in the original PR post. This loop does not actually drop anything anymore. Before, it would drop the `Moved` entries in the bin vector, but now we don't have these. So all this does currently is check that bins are either `null` or a reference to the shared `Moved`. However, this should be the case anyways as outline by the comment above the loop.

src/raw/mod.rs
domenicquirl: Hm, this is done in `drop_bins` but not here, I wonder why. If we decided against the checks (see also above), would we even need to swap anything or would `bins` just get dropped automatically?

jonhoo:
Hmm, okay, stepping out from the sub-comments a bit, I think the loop should be replaced with just a swap of `.bins` with an empty vector. Then,`if cfg!(debug_assertions)`, walk the taken vector and check that all the entries are either null or `Moved`. Then, `drop()` the taken vector _before_ we ultimately drop the shared `Moved`.

domenicquirl:
It was, except I didn't know about the possibility of using `cfg!` to still do the check in testing. That's really nice! This should be ready then 😊

src/raw/mod.rs
```diff
@@ -29,6 +72,100 @@ impl<K, V> Table<K, V> {
...
+                assert_eq!(t, for_table);
+            }
+        }
+        // return a shared pointer to BinEntry::Moved
+        self.moved.load(Ordering::SeqCst, guard)
+    }
+
+    pub(crate) fn find<'g, Q>(
+        &'g self,
+        bin: &BinEntry<K, V>,
```
jonhoo: Come to think of it, now that this method is on `Table`, can't we just pass the `bini` here instead? I guess taking the `&BinEntry` directly saves us a load potentially?

src/raw/mod.rs
```diff
@@ -76,26 +217,52 @@ impl<K, V> Table<K, V> {
...
-                continue;
+        // since BinEntry::Nodes are either dropped by drop_bins or transferred to a new table,
+        // all bins are empty or contain a Shared pointing to shared the BinEntry::Moved (if
+        // self.bins was not replaced by drop_bins anyway)
+        let bins = Vec::from(std::mem::replace(&mut self.bins, vec![].into_boxed_slice()));
+
+        // when testing, we check the above invariant. in production, we assume it to be true
+        if cfg!(debug_assertions) {
+            for bin in bins.iter() {
+                let bin = bin.swap(Shared::null(), Ordering::SeqCst, guard);
```
jonhoo: This can now just be a `.load`

src/raw/mod.rs
```diff
@@ -76,26 +217,52 @@ impl<K, V> Table<K, V> {
...
             // safety: we have mut access to self, so no-one else will drop this value under us.
-            let bin = unsafe { bin.into_owned() };
-            if let BinEntry::Moved(_) = *bin {
-            } else {
-                unreachable!("dropped table with non-empty bin");
-            }
+            let moved = unsafe { moved.into_owned() };
+            drop(moved);
+        } else {
+            unreachable!("self.moved is initialized together with the table");
```
jonhoo: I'd probably just do
```rust
assert!(!moved.is_null(), "self.moved is initialized together with the table")
```
above and avoid the conditional.

jonhoo:
This is looking really good now ­— just a few more comments :sweat_smile:

src/raw/mod.rs
domenicquirl: We could, but both in `get_node` and in `find` (the places that call `find`) there is a `null` check before. In `find`, we also `deref()` and `match` the `bin` to decide whether we found a `Node` or need to follow a `Moved` to a next table. 

This implementation is has a double `match` anyways due to the recursive call to `find`, but if we change this to `bini` we would `load` and `deref` the bin anyways (for the `match`), only to then pass `bini` and `load` and `deref` again. 

I don't see a huge advantage in this. If anything, one would want to `find` _only with_ hash and key. But `hash()` is on the map and doing this is basically what `get_node` is for, though I agree that it would be the nicer interface for a method called `find`. But maybe this is more due name and less due to function?

jonhoo:
Beautiful, thank you! :tada:

PR #57, opened 2020-02-03
Match up documentation with that from std::collections::HashMap

Arthamys:
This PR updates the documentation to match that of std::HashMap, along with simple examples.

The examples make use of `flurry::epoch::pin()`. I don't know if this is still desirable, or if the documentation should make use of `HashMap::guard()`.


jonhoo:
This is great! I'll take a look through once I get the chance.

Two high-level comments:
 - We probably want to use @cuviper's new `HashMapRef` type in most of the examples, as it avoids the need to produce a guard manually. Just `map.pin().get(key)` for example. That should tidy up a lot of the example code I think.
 - When there are comments in the documentation, you need to enclose them in `/* */` so that they do not show up in the _actual_ documentation :)

cuviper:
Oh, the joys of concurrent data structures -- "look ma, no `mut`!"

src/lib.rs
```diff
@@ -55,12 +55,17 @@
 //! operation. When possible, it is a good idea to provide a size estimate by using the
 //! [`with_capacity`](HashMap::with_capacity) constructor. Note that using many keys with
 //! exactly the same [`Hash`](std::hash::Hash) value is a sure way to slow down performance of any
-//! hash table. /* TODO: dynamic load factor */
+//! hash table.
 //!
-//! /* TODO: set projection */
-//! /* TODO: frequency map through computeIfAbsent */
-//! /* TODO: bulk operations like forEach, search, and reduce */
+/*
```
jonhoo: I'm not sure Rust will let you get away with breaking up a crate-level doc comment this way? You may have to do it the same way we did in the original version, with `/* */` around each individual thing, or with `//! /*` to start and `//! */` to end.

Arthamys: The crate-level documentation is fine with this approach.
The reason I changed the comments is that the `TODO` blocks were displayed in the generated docs. I hope I'm not getting it wrong that you want the TODOs to be hidden from the user facing documentation, right ?
I joined a screenshot of the doc comment and the render to illustrate
2020-02-04-002240_1920x1080_scrot

jonhoo: Oh, interesting, I was completely wrong then! Ignore me :sweat_smile: The behavior we want is that the TODOs are hidden from the docs, which it looks like your approach does!

src/lib.rs
```diff
@@ -76,13 +81,15 @@
...
 //!
 //! This map usually acts as a binned (bucketed) hash table.  Each key-value mapping is held in a
-//! `BinEntry.  Most nodes are of type `BinEntry::Node` with hash, key, value, and a `next` field.
+//! `BinEntry`.  Most nodes are of type `BinEntry::Node` with hash, key, value, and a `next` field.
 //!  However, some nodes are of type `BinEntry::Moved`; these "forwarding nodes" are placed at the
 //!  heads of bins during resizing. The Java version also has other special node types, but these
 //!  have not yet been implemented in this port. These special nodes are all either uncommon or
 //!  transient.
-//! /* TODO: TreeNodes, ReservationNodes */
+/*
```
jonhoo: We probably want to preserve the paragraph break here, which I don't _think_ this will do?

src/lib.rs
```diff
@@ -181,8 +191,10 @@
 //! occurring at threshold is around 13%, meaning that only about 1 in 8 puts check threshold (and
 //! after resizing, many fewer do so).
 //! */
```
jonhoo: This probably also needs to be changed then?

src/map.rs
```diff
@@ -160,15 +160,35 @@ where
...
+    /// ```
     pub fn new() -> Self {
         Self::default()
     }
 
-    /// Creates a new, empty map with an initial table size accommodating the specified number of
-    /// elements without the need to dynamically resize.
-    pub fn with_capacity(n: usize) -> Self {
-        Self::with_capacity_and_hasher(n, crate::DefaultHashBuilder::default())
+    /// If `capacity` is 0, the hash map will not allocate
```
jonhoo: This looks like a stray line?

src/map.rs
```diff
@@ -160,15 +160,35 @@ where
...
+    /// Creates an empty `HashMap` with the specified capacity.
+    ///
+    /// The hash map will be able to hold at least `capacity` elements without
+    /// reallocating. If `capacity` is 0, the hash map will not allocate.
+    ///
```
jonhoo: We may want to mention here that our `with_capacity` (as opposed to `std::collections::HashMap` I believe) does not _guarantee_ that we won't resize if `capacity` elements are inserted. In general, the map will be resized once a certain amount of collisions are detected, so with particularly bad key distributions, we may resize sooner.

src/map.rs
```diff
@@ -250,6 +279,15 @@ where
     /// Warning: `hash_builder` is normally randomly generated, and is designed to allow the map
     /// to be resistant to attacks that cause many collisions and very poor performance.
     /// Setting it manually using this function can expose a DoS attack vector.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use flurry::{HashMap, DefaultHashBuilder};
+    ///
+    /// let map = HashMap::with_capacity_and_hasher(10, DefaultHashBuilder::default());
```
jonhoo: We could actually use the `RandomState` hasher in our example here. Using the default hasher seems a little odd, since it is the default after all :p

src/map.rs
```diff
@@ -277,10 +315,25 @@ where
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use flurry::HashMap;
+    ///
+    /// let map = HashMap::new();
+    /// map.pin().insert(1, "a");
+    /// assert_eq!(map.pin().contains_key(&1), true);
+    /// assert_eq!(map.pin().contains_key(&2), false);
```
jonhoo: This may be a good place to demonstrate that you can re-use `map.pin()`:
```rust
let mref = map.pin();
mref.insert(1, "a");
assert_eq!(mref.contains_key(&1), true);
assert_eq!(mref.contains_key(&2), false);
```

src/map.rs
```diff
@@ -342,14 +395,27 @@ where
...
     ///
-    /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
-    /// form must match those for the key type.
+    /// # Examples
+    ///
+    /// ```
+    /// use flurry::HashMap;
+    ///
+    /// let map = HashMap::new();
+    /// map.pin().insert(1, "a");
```
jonhoo: Same here, let's re-use `.pin`. And also in other examples.

src/map.rs
```diff
@@ -368,12 +434,27 @@ where
     }
 
     #[inline]
-    /// Obtains the value to which `key` is mapped and passes it through the closure `then`.
-    ///
+    /// Apply `then` to the mapping for `key` and get its result.
```
jonhoo: I think we may want to remove this method altogether now, but that's for a different PR.

src/map.rs
```diff
@@ -446,15 +527,59 @@ where
...
+    /// assert_eq!(map.pin().insert(37, "a"), None);
+    /// assert_eq!(map.pin().is_empty(), false);
+    ///
+    /// map.pin().insert(37, "b");
+    /// assert_eq!(map.pin().insert(37, "c"), Some(&"b"));
+    /// assert_eq!(map.pin().get(&37), Some(&"c"));
+    /// ```
     ///
-    /// The value can be retrieved by calling [`HashMap::get`] with a key that is equal to the original key.
+    /// # Notes
```
jonhoo: I wonder if we should move this note up to the type-level docs rather than having it hidden away on only some methods. What do you think?

src/map.rs
Arthamys: You're right, this will better illustrate the function

src/map.rs
Arthamys: I agree, the notion of pinning transpires through the whole API, so might as well put in a place that will be more prominent in the docs.
Should I also remove the notes on the `iter()`, `keys()` and `values()` methods ?

jonhoo: Yes, I think we should have it _just_ on `HashMap`.

src/map.rs
```diff
@@ -55,7 +55,17 @@ macro_rules! load_factor {
 
 /// A concurrent hash table.
 ///
-/// See the [crate-level documentation](index.html) for details.
+/// Flurry uses an [`Guards`] to control the lifetime of the resources
+/// that get stored and extracted from the map.
+///
+/// [`Guards`] are acquired through the [`epoch::pin`], [`HashMap::pin`] and
+/// [`HashMap::guard`] functions.
+///
```
jonhoo: These can probably be joined into one paragraph.

src/map.rs
```diff
@@ -283,9 +301,11 @@ where
     /// # Examples
     ///
     /// ```
-    /// use flurry::{HashMap, DefaultHashBuilder};
+    /// use flurry::HashMap;
+    /// use std::collections::hash_map::RandomState;
     ///
-    /// let map = HashMap::with_capacity_and_hasher(10, DefaultHashBuilder::default());
+    ///let s = RandomState::new();
```
```suggestion
    /// let s = RandomState::new();
```

jonhoo:
I'm working on fixing the failing address sanitizer stuff in https://github.com/jonhoo/flurry/pull/59

jonhoo:
Excellent, thank you so much!

PR #61, opened 2020-02-17
Add more tests

Stupremee:
This PR adds several tests to increase the test coverage and more benchmarks.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/61)


jonhoo:
(CI failure is due to `miri` missing on `2020-02-21` nightly: https://rust-lang.github.io/rustup-components-history/)

Stupremee:
Should I fix the ci issue?

tests/basic_ref.rs
```diff
@@ -368,33 +271,41 @@ fn drop_value() {
...
+    let cloned_map = map.clone();
     assert_eq!(map.len(), cloned_map.len());
-    assert_eq!(map, cloned_map);
+    assert_eq!(&map, &cloned_map);
     assert_eq!(cloned_map.len(), 0);
 }
 
 #[test]
-// Test that same values exists in both refs (original and cloned)
+// TODO: Should this test really fail?
```
Stupremee: What should we do with this test?

jonhoo:
> Should I fix the ci issue?

No, just leave it -- it'll likely be fixed tomorrow-ish.

tests/basic_ref.rs
jonhoo: Ah, I think that test is _supposed_ to hold an `Arc<HashMap>`, not just a `HashMap`, and that the clone is supposed to just clone the `Arc`, not the entire `HashMap`. With that, the original test would be correct.

Stupremee: If I change the "normal" (`HashMap` instead of `HashMapRef`) to use Arc, it fails with the same error.

jonhoo: What error is that? Are you using `Arc::clone` to clone the `Arc<HashMap>` (not just calling `.clone()`)?

Stupremee: If the original test uses `Arc`, it fails with the same error.

Stupremee: Yep I'm using `Arc::clone`

jonhoo: Right, but what error is that? Also, when changing it to use `Arc`, you also need to change the line that does `.clone()` to instead do `Arc::clone(&map)`.

Stupremee:
```rs
---- clone_map_filled stdout ----
thread 'clone_map_filled' panicked at 'assertion failed: `(left != right)`
  left: `{"FooKey": 0, "NewItem": 100, "BarKey": 10}`,
 right: `{"FooKey": 0, "NewItem": 100, "BarKey": 10}`', tests/basic.rs:491:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
That's the error

jonhoo: Oh oh oh, I think I understand what you mean now. So, the original test is testing that if you clone the map, you end up with two distinct copies, and that modifying one copy does not modify the other. That test is correct. With `HashMapRef`, the semantics are different: cloning a `HashMapRef` gives you a new `HashMapRef` _to the same map_. Those are the semantics we want (I believe), so the test for ref will necessarily have to be different. Specifically, in ref, it should assert that the two clones of the map are indeed equal after the modification.

Stupremee: Ahh okay. Thank you :+1:

tests/basic_ref.rs
```diff
@@ -289,9 +288,10 @@ fn clone_map_filled() {
     assert_eq!(map.len(), cloned_map.len());
     assert_eq!(&map, &cloned_map);
 
-    // test that we are not mapping the same tables
+    // test that both maps are equal,
+    // because the ref and the cloned ref, point to the same map
     map.insert("NewItem", 100);
-    assert_ne!(&map, &cloned_map);
+    assert_eq!(&map, &cloned_map);
```
jonhoo: Maybe call these `map_ref` and `cloned_map_ref` to make it clear?

Stupremee:
Are there some must-have test / benchmarks that you would like to have in Flurry?
Because I wasn't able to find any missing tests from the Jdk.
I will just add some benchmarks.

jonhoo:
The tests that I know of are the [JDK tests](https://hg.openjdk.java.net/jdk/jdk13/file/tip/test/jdk/java/util/concurrent/ConcurrentHashMap), the [JSR166 tests](https://github.com/jonhoo/flurry/blob/master/jsr166/test/ConcurrentHashMapTest.java), and the [`hashbrown` tests](https://github.com/rust-lang/hashbrown/tree/master/tests). Ideally we'd have pretty much all the tests from all those three.

Stupremee:
I think all tests are implemented that are compatible with flurry.
I think I will leave the benchmarks as they are.

domenicquirl:
Thanks for working on this! A few comments from looking through the diff:
- You added the `OneBucketHasher` from `basic.rs` to `basic_ref.rs` to use it there as well, and there is also `ZeroHasher` in `hasher.rs`. All of these hash everything to `0`. It would probably make sense to have this only once and shared it everywhere it is needed.
- It looks like you simplified uses of `map.pin()` in most tests. @jonhoo did you look at this in your previous review?
  - For some methods, such as `insert_and_remove`, this is probably okay since they also only used one `pin()` before.
  - Other methods use several pins, and we have to be careful to not involuntarily exclude testing that using multiple `HashMapRef`s works as intended. For example in `update`, we want the updates to be visible if we use a different Ref later. I think we should keep testing that.
  - The `XYZ_equal` methods previously tested `HashMapRef` equality against `HashMapRef` itself and against `HashMap`. Since these equality traits are added together with `HashMapRef`, I think the test should continue to check this.
- Concurrent tests: 
  - When setting up a map at the start of a concurrent test, e.g. [here](https://github.com/jonhoo/flurry/blob/45b956ed1507fac69a0f7fe0ec653b64d862c8af/tests/concurrent_contains.rs#L16), if we use a guard and leave it around afterwards, then this guard will pin an epoch for the entire duration of the test. To avoid this, the guard can be put inside of a local scope, like here:
https://github.com/jonhoo/flurry/blob/master/tests/basic.rs#L253-L262
  - Also note that this uses `map.guard()` instead of `epoch::pin()`, which we could change while we're at it.
  - This one might be me not exactly keeping up with everything that happend here (if so, sorry 😅 ), but I didn't see new versions of some of the removed concurrency tests, for example `concurrent_remove`. If there isn't a new implementation, I would just keep the old tests so we don't loose anythin.
- Lastly, there is a typo [here](https://github.com/jonhoo/flurry/blob/45b956ed1507fac69a0f7fe0ec653b64d862c8af/tests/hasher.rs#L43), but that'll be quick to change 😊

domenicquirl:
Wonderful! Don't know why one of the CI targets failed, but all of my comments have been addressed with the new commits, so I'm happy to approve this now.

jonhoo:
The miri failure is because miri was broken on the last nightly. I think the newest one has it again though, so I'll re-run.

jonhoo:
:tada: Thank you!

PR #63, opened 2020-02-20
Also get coverage from doctests

jonhoo:


jonhoo:
I had to make one kind of sad change due to https://github.com/xd009642/tarpaulin/issues/344, but it's fine.

PR #66, opened 2020-02-26
Remove Clone for with_capacity with simple presize

jonhoo:
Follow-on work from #64

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/66)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -176,6 +176,30 @@ impl<K, V> HashMap<K, V, crate::DefaultHashBuilder> {
     pub fn new() -> Self {
         Self::default()
     }
+
+    /// Creates an empty `HashMap` with the specified capacity.
+    ///
+    /// The hash map will be able to hold at least `capacity` elements without
+    /// reallocating. If `capacity` is 0, the hash map will not allocate.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use flurry::HashMap;
+    /// let map: HashMap<&str, i32> = HashMap::with_capacity(10);
+    /// ```
+    ///
+    /// # Notes
+    ///
+    /// There is no guarantee that the HashMap will not resize if `capacity`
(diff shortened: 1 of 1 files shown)


src/map.rs
```diff
@@ -382,6 +436,47 @@ impl<K, V, S> HashMap<K, V, S> {
...
+            // round the requested_capacity to the next power of to from 1.5 * size + 1
+            // TODO: find out if this is neccessary
+            let size = size + (size >> 1) + 1;
+
+            std::cmp::min(MAXIMUM_CAPACITY, size.next_power_of_two())
+        } as usize;
+
+        // sanity check that the map has indeed not been set up already
+        assert_eq!(self.size_ctl.load(Ordering::SeqCst), 0);
+        assert!(self.table.load(Ordering::SeqCst, &guard).is_null());
```
domenicquirl: Should we add a short note here about what would need to be done if at some point we allow a non-zero minimum capacity? Just something that makes clear that if `with_capacity` fails the `size_ctl` assertion it's not because it _has_ to always be 0, but because it is not handled.

jonhoo: I'm not sure I follow? This method should only ever be called if the map has not yet been allocated, at which point `size_ctl` should indeed always be zero. Or am I missing something?

domenicquirl: `size_ctl` is _currently_ always zero, but it doesn't have to be. In general, it stores the minimum capacity allocated for the table. This value is checked in `try_presize` and overrides `DEFAULT_CAPACITY` in `init_table` if `> 0`.

I'm not saying we have to implement this behaviour for `presize` now, I would assume `size_ctl` to stay `0`. But if someone were to change this in the future, this assertion makes it look a lot like it _must_ be `0` if we use `presize`. It's not _that_ important, I already approved the changes themselves 😉

PR #67, opened 2020-02-27
Fix up inline annotations.

jonhoo:
Also removes `get_and`, because it's pretty useless now.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/67)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1167,37 +1167,6 @@ where
         unsafe { v.as_ref() }
     }
 
-    #[inline]
-    /// Apply `then` to the mapping for `key` and get its result.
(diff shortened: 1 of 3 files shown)


twe4ked:
Out of interest, how did you decide what things to `inline` and what things to `inline(never)`? If this is a huge topic feel free to ignore 😅!

jonhoo:
@twe4ked I mostly just followed what `hashbrown` does, because I know they've thought a bunch about this. The `inline(never)` was there because a bunch of our methods contain calls to `transfer`, but it should be so rare (and so expensive), that inlining is probably never worth it.

PR #68, opened 2020-03-02
Expand HashSet impls with more methods/traits

jonhoo:
This also aligns HashSet more with HashMap (such as on bounds).

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/68)


twe4ked:
Nice changes, that would have taken a while with PR feedback. Now that I see the changes they're obvious 😛. Thanks!

jonhoo:
Hehe, yeah, I figured it might be easier to just do them than trying to walk through them one by one :p Hopefully you'll still find it useful to see them written out though!

PR #72, opened 2020-03-16
Implement TreeBins

### LINKED ISSUE #13: Implement the tree-bin optimization
The Java code modified large bins to be trees instead of lists. We should do the same. See the [`TreeNode` optimization] and the [implementation notes][tree-impl] on that optimization for much more detail.

  [`TreeNode` optimization]: https://github.com/jonhoo/flurry/blob/d3dae0465b37b7f12c4f0d58a16f36fb1d8c1596/jsr166/src/ConcurrentHashMap.java#L327-L339
  [tree-impl]: https://github.com/jonhoo/flurry/blob/d3dae0465b37b7f12c4f0d58a16f36fb1d8c1596/jsr166/src/ConcurrentHashMap.java#L413-L447

domenicquirl:
This PR ports an optimization from the Java code to flurry which replaces linear bins with bins containing ordered trees for bins containing many elements (in case of many hash collisions). 

TreeBins and TreeNodes are introduced as structs with all associated methods, including the red-black tree implementation and methods to read from or insert into a tree bin. Handling for tree bins is added to all methods of `HashMap` that are sensitive to bin content/the type of bins, including iterators. Documentation is provided for the implementation, in particular the crate-level implementation notes are updated.

Documentation changes may also include minor modifications to where I noticed things to be out-of-date. These are small across the board, notably I included some information on `HashSet`s in the crate-level documentation, since they were added with #19 (which means we should also probably re-purpose #17 to address missing methods or `HashSetRef` only, or close it and move these to a new issue).

Since I did not write most of the methods of `HashMap` that I had to touch, please review these particularly vigilantly. In particular the iterator code, since this was implemented in my absence.

Closes #13.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/72)


domenicquirl:
Hm, is something up with CI? I wanted to start this as a draft to check ASAN, but CI seemed unhappy to start so I marked this as ready. Still doesn't seem to get going :/

jonhoo:
Hmm, could you try rebasing on master and force-pushing?

domenicquirl:
Aha! There seems to be some remaining issues, but at least I can address them now!

domenicquirl:
Actually, I'm a bit stumped by this. ASAN complains about the part of `transfer` where new tree nodes are created for both the [high](https://github.com/domenicquirl/flurry/blob/treebin-optimization/src/map.rs#L960) and the [low](https://github.com/domenicquirl/flurry/blob/treebin-optimization/src/map.rs#L943) bin. The fact that it's this spot specifically makes me think it has to do with the case that these nodes may be discarded if the new bin is no longer big enough to be a tree bin (since it was split in two). 

However, I am aware that in this case these nodes need to be discarded and even did so directly in order to not go through the epoch garbage collection. This should be handled by the code [here](https://github.com/domenicquirl/flurry/blob/treebin-optimization/src/map.rs#L985-L996) and in the corresponding case below. Can you maybe take a look at that? Turning the nodes back into owned references happens locally, but I thought that should still drop them, or is that not the case?

domenicquirl:
Managed to figure it out in my remaining time this morning before I have to leave. It _was_ the re-used bins that leaked memory, just had to account for the fact that additionally, values are always re-used in the new table and cannot be dropped both with the old and the new bin.

miri seems to be unavailable again, but the other checks all pass now, so this should now actually be ready for review 😊

jonhoo:
https://github.com/jonhoo/flurry/pull/73 should fix miri not running

domenicquirl:
Yeah, that's why I said I'll do a few batches ^^
No worries though, that's what this is for after all. And I'd rather read some more comments and get this right than miss something.

I didn't push any changes yet, I still need to rebase in the fix for miri. Will push after the next set of updates so you can see what has changed. But that test thing took so long that I didn't want to get drawn into fixing more stuff in case ASAN complains or something 😅 

The code for the different kind of drop is already methods, this should be good now 👍

jonhoo:
Let's hope that reviewable deals better with rebases than GitHub does. That's the primary reason why I'm trying it out :sweat_smile:

domenicquirl:
The comment about the efficiency of `defer_destroy` may also apply to your concern regarding the use of `defer_unchecked`, given that the former also creates a closure which is then handled by the latter. I'll stop here for today though, we can come back to this later.

domenicquirl:
It is _damn_ confusing to merge in an explicit drop of a node whose creation I removed ^^
Missed some errors because apparently both my IDE plugin and `cargo test` don't use all features? Gimme a sec.

domenicquirl:
Checks passing now! 🎉 
It's only coverage that's failing. Admittedly, there aren't that many tests for the TreeBins, but the red-black tree stuff is checked by `check_invariants` in testing (which I don't think counts as an actual test?). The `node.rs` drop is probably the fault of my changes, `set.rs` has to be the merged-in changes though?

Would be nice to have some more tests for the more delicate parts of TreeBins, but these are also kind of hart to trigger. For example, it would be nice to have a test which always triggers a writing thread waiting for the additional lock. How do you want to proceed? I can't really tell how much of this is on me and what was merged in.

domenicquirl:
Looks like it slightly helps `node.rs`, but doesn't really help in general :/

src/map.rs
```diff
@@ -3356,6 +3356,75 @@ mod tree_bins {
...
+                    if let Some(v) = map.get(&key, guard) {
+                        criterion::black_box(v);
+                    }
+                }
+            }));
+        }
+
+        // in the end, join all threads
+        for handle in handles {
+            if let Err(_) = handle.join() {
```
jonhoo, line 3422: isn't this just an `unwrap`?

src/map.rs
```diff
@@ -3356,6 +3356,75 @@ mod tree_bins {
...
+                BinEntry::Node(_) => panic!("bin was not correctly treeified -- is Node"),
+                BinEntry::TreeNode(_) => panic!("bin was not correctly treeified -- is TreeNode"),
+            }
+
+            guard.flush();
+            drop(guard);
+        }
+        // then, spin up lots of reading and writing threads on a small range of keys
+        const NUM_THREADS: usize = 100;
+        const NUM_REPEATS: usize = 1000;
```
jonhoo, line 3386: This number is probably too low.

src/map.rs
```diff
@@ -3356,6 +3356,75 @@ mod tree_bins {
...
+                BinEntry::Moved => panic!("bin was not correctly treeified -- is Moved"),
+                BinEntry::Node(_) => panic!("bin was not correctly treeified -- is Node"),
+                BinEntry::TreeNode(_) => panic!("bin was not correctly treeified -- is TreeNode"),
+            }
+
+            guard.flush();
+            drop(guard);
+        }
+        // then, spin up lots of reading and writing threads on a small range of keys
+        const NUM_THREADS: usize = 100;
```
jonhoo, line 3385: This number probably doesn't need to be that high.

jonhoo:
Looking through the coverage diff, it doesn't actually look too bad. I think one thing that's not being tested is deletes, which may cause untreefication and such. Since the set of keys is small, I think this also maybe mostly ends up testing value overwrites rather than tree restructuring (like rotations), so maybe the keyspace actually needs to be a decent amount larger (and have deletes)?

domenicquirl:
But do I read the individual file view correctly in that `contended_lock` was not hit by any test? That would be what we want to trigger...
I (locally) added delete threads, as well as an increased key range, which should help with that. We may run into an issue with test execution times here however:
```rust
const NUM_THREADS: usize = 20;
const NUM_REPEATS: usize = 1000;
const NUM_KEYS: usize = 1000;
```
is the last configuration that runs in a somewhat reasonable time. Both running 1000 repeats with 10000 different keys and 10000 repeats with 1000 different keys takes a minute plus. Both on 10000 took a whopping 10 minutes to run.

jonhoo:
Yeah, I feel like we _really_ want loom here :sweat_smile:
Until we do, all we can hope for is a rough approximation for the contention cases I think sadly.

domenicquirl:
That seems to have done quite a lot still. It's only about half the impact for `node.rs` now (still -40% for `set.rs`, but I don't know which changes these are). Overall impact is down to -8.5 from -14.

I think `contended_lock` is still not being hit? I'm not quite familiar with the interface. If so, that would be a bit sad, but maybe we really need loom for this.

Any more comments or ideas? Otherwise I think we can finally land this!

jonhoo:
The stuff in `set.rs` I think the coverage tool is just confused about, and I would ignore it.

It also isn't hitting things like a bin being removed just as someone tries to insert into it, but I think that's okay, and will require loom. Some of the coverage is also just incorrectly marked as untested, probably due to limitations in the coverage sampler (like `unwrap` being marked as uncovered because we don't explore the panic branch). We also run into some limitations around all the buckets hashing to 0, which means that they all get moved _together_ in the case of a resize, but that too I think is okay. Similarly, our comparison by hash always returns `==`, but that is the most interesting case, so that's fine. It looks like the drop code for `BinEntry::Tree` [isn't hit], which is odd? `contended_lock` doesn't seem to be hit, you're right, which is really sad. Not sure how to force that beyond what we're already doing. Maybe fewer writing threads would help, since it lets the reader threads be more aggressive. Might also be we just don't hit it on CI, since it has fewer cores, and likely the many threads this test has can't all run at once. The `if p.is_null()` clause in `find_or_put_tree_val` also appears to never be hit? Same with the [drop closure in `defer_drop_without_values`](https://codecov.io/gh/jonhoo/flurry/pull/72/diff?src=pr&el=tree-more#D10-927) for some reason?

Overall, I'm pretty happy though. We will want loom for this regardless. Maybe try to get coverage for the drop, and try fewer writers (both insert and remove) but with the same number of readers, and then we call it good?

jonhoo:
Looks like `contended_lock` still isn't hit :'(

domenicquirl:
The `find_or_put_tree_val` case makes sense to me, since the tree bin should be created from an existing linear bin, in which case `TREEIFY_THRESHOLD` elements are present initially, including a root. It's one of these cases where it's unclear whether `root` can even ever be null.

Similarly, I would assume that the drop method is not hit by `transfer` because it can always re-use the existing bin (because the high bin is empty). It _should_ be hit by `untreeify` which is covered by the other tests, so is weird that the closure wouldn't be hit. Latest CI run on the new commit _does_ hit the closure.

`contended_lock` still isn't, yeah, but I don't know how else to force that. With another slight improvement I am in favour of calling it good and waiting for loom to cover this...

jonhoo:
Awesome!

PR #73, opened 2020-03-17
Proper miri setup + pin miri nightly

jonhoo:
@RalfJung [suggested](https://github.com/jonhoo/flurry/commit/7bd22c62eac86a8e0d3426b5a4c790e9850e5843#r37880279) we use `cargo miri setup`, so we do. I've also pinned the nightly to one that has miri since the current nightly does not (https://github.com/rust-lang/rust/issues/70055).

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/73)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,9 +47,12 @@ jobs:
    pool:
      vmImage: ubuntu-latest
    steps:
+     - bash: |
+         echo '##vso[task.setvariable variable=nightly]nightly-'$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/miri)
+       displayName: "Determine latest miri nightly"
      - template: install-rust.yml@templates
        parameters:
-         rust: nightly
+         rust: $(nightly)
          components:
            - miri
      - script: cargo miri setup
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -52,8 +52,13 @@ jobs:
          rust: nightly
          components:
            - miri
-     # ignore leaks due to https://github.com/crossbeam-rs/crossbeam/issues/464
-     - bash: yes | cargo miri -Zmiri-ignore-leaks test
+     - script: cargo miri setup
+       displayName: cargo miri setup
+     # ignore leaks due to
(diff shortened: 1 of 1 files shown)


RalfJung:
The Miri README also explains how to get the latest nightly where Miri works. :)

```
MIRI_NIGHTLY=nightly-$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/miri)
echo "Installing latest nightly with Miri: $MIRI_NIGHTLY"
rustup set profile minimal
rustup default "$MIRI_NIGHTLY"
```

jonhoo:
@RalfJung Ah, good to know! That _may_ be tricky to combine with Azure templates though, since it sort of wants to know things ahead of time. But I'll see what I can come up with!

azure-pipelines.yml
```diff
@@ -47,11 +47,13 @@ jobs:
...
+     - bash: |
+         echo '##vso[task.setvariable variable=Nightly]nightly-'$(curl -s https://rust-lang.github.io/rustup-components-history/x86_64-unknown-linux-gnu/miri)
+       displayName: "Determine latest miri nightly"
+       name: LatestMiri
      - template: install-rust.yml@templates
        parameters:
-         # fix nightly since miri breaks there frequently
-         # should probably be bumped occasionally
-         rust: nightly-2020-03-16
+         rust: $(LatestMiri.Nightly)
```
RalfJung: oO WTF what is this.^^ I thought one would just execute the commands I gave as a shellscript. But I also never used Azure CI.

If this works, do you think it is worth putting that into the Miri doc as well?

jonhoo: This is the awfulness that is [Azure Pipelines Templates](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops). Unless people are using templates, they shouldn't need this syntax. Here I am using the (very handy) crate-ci/azure-pipelines [install-rust template](https://github.com/crate-ci/azure-pipelines/blob/995c1b8e848994c775ac8349a9801d7dee3a4193/install-rust.yml#L9-L14).

Issue #58, opened 2020-02-04
Expose non-replacing `insert`


jonhoo:
By calling `put` with `no_replacement = true`, we can provide a version of `insert` that does _not_ update the value if the key is already present. I can imagine that being a relatively useful method to provide. I'm not sure what to call it though? `std::collections::HashMap` does not have an equivalent, so we're on our own here.

llitz:
noreplace comes to mind but that's too cheesy

Perhaps insertifnull

domenicquirl:
A lot of the APIs which deal with concurrency use `try_` for methods that depend on state shared between multiple threads. For example, a lot of the locks have a `try_lock()` method which returns `false` if the lock is already held by another thread. Such methods can also be found in places where an operation might not succeed, such as `try_from` for fallible conversions.

 As I assume we have a similar use case for the proposed method (i.e. inserting if no value for the given key is currently stored by a different thread) and it is definitely a case of an operation that might not happen depending on the state of the map, I suggest following this naming scheme and calling the method `try_insert`.

jonhoo:
Oh, wow, yes, `try_insert` is a great name!

llitz:
try_insert feels weird compared to the other APIs. For example, we do a try_lock() and it fails, the lock wasn't acquired. Now do a try_insert(), the insert failed but the data is there anyway.

It feels the wrong name to call based on what I usually expect from try, something that can fail. If the data is already there, is it a failure?

domenicquirl:
The proposal only mentions checking if an entry already exists for the given key. The map may still contain a value different from what you try to insert, so in this case you "failed" to store _your_ provided value in the map. So while there might be data in the map after a such a failed `try_insert`, it is not generally the data you wanted it to contain. That's how I think of this at least.

We can also discuss whether we should check the already stored value against the given value and what to return in case they match.

llitz:
Hmmm you are right it makes a lot of sense to use try_insert from considering something else can be there.

Checking the value may have a high cost, but could be necessary to decide what to return.

jonhoo:
I don't think we should check for equality of the value.

One alternative is to rename the _current_ `insert` to `upsert`, which is a short-form often used for "update or insert", and then call this method `insert`. But that's _probably_ more confusing, since we then would not match the `std` API.

domenicquirl:
[...]

PR #74, opened 2020-03-22
Implemented `HashMap::try_insert`

GarkGarcia:
Follow up on #58.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/74)


src/map.rs
```diff
@@ -1348,6 +1348,38 @@ where
         self.put(key, value, false, guard)
     }
 
+    #[inline]
```
domenicquirl: Nit: We recently cleaned up all `#[inline]` annotations to be directly above the method signature (below the comment) with https://github.com/jonhoo/flurry/commit/c326f681cd729ad01cc109f450610cbbd408f059. It would be nice if we kept that consistent now

src/map.rs
```diff
@@ -1348,6 +1348,38 @@ where
...
+    /// Tries to insert a key-value pair into the map.
+    ///
+    /// If the map did not have this key present, [`None`] is returned and the
+    /// key-value pair is inserted into the map.
+    ///
+    /// If the map did have this key present, and the old value is returned.
+    /// Neither the key or value are updated; this matters for types that can
+    /// be `==` without being identical. See the [std-collections
```
domenicquirl: I would maybe change `==` to refer to `Eq` here

src/map_ref.rs
```diff
@@ -154,6 +154,14 @@ where
         self.map.insert(key, value, &self.guard)
     }
 
+    #[inline]
```
domenicquirl: Nit: Same here with `#[inline]`

src/map.rs
```diff
@@ -1348,6 +1348,38 @@ where
...
+    /// let map = HashMap::new();
+    /// let mref = map.pin();
+    ///
+    /// mref.insert(37, "a");
+    /// assert_eq!(mref.try_insert(37, "b"), Some(&"a"));
+    /// assert_eq!(mref.try_insert(42, "c"), None);
+    /// assert_eq!(mref.get(&37), Some(&"a"));
+    /// assert_eq!(mref.get(&42), Some(&"c"));
+    /// ```
+    pub fn try_insert<'g>(&'g self, key: K, value: V, guard: &'g Guard) -> Option<&'g V> {
```
domenicquirl: Regarding the return type and https://github.com/jonhoo/flurry/issues/58#issuecomment-602286897, I think `Result<(), &'g V>` is fine. The existing `try_XYZ` implementations differ in what they return as `Ok(_)`, depending on you would expect a return value from the "regular" version of these operations. For example, `Mutex::try_lock` returns `Ok(Guard)`, where `Guard` can then be used to access the object behind the Mutex. But `HashMap::try_reserve` returns `Ok(())`, since you'd probably not expect a return when reserving some space.

In our case, the regular version of `try_insert` is simply `insert`, where the key is not present in the map yet. The return of `insert` in that case is `None`, which is about as informative as `()` if you know the key can't have been there before ^^

src/map.rs
GarkGarcia: No problem, I'll fix this.

src/map.rs
GarkGarcia: The same wording is used in the documentation of `flurry::HashMap::insert`, should I change it there too?

src/map.rs
GarkGarcia: Seems reasonable.

jonhoo: See also my argument in https://github.com/jonhoo/flurry/issues/58#issuecomment-602568241 for having `Err` be:
```rust
struct TryInsertError {
    current: &T,
    new: T,
}
```

domenicquirl: This would still necessitate changes to `put`, right? Because ownership of the new value is transferred to it (so it can create a new node) and we would need to get it back.

src/map.rs
```diff
@@ -1348,6 +1348,40 @@ where
         self.put(key, value, false, guard)
     }
 
+    /// Tries to insert a key-value pair into the map.
+    ///
+    /// If the map did not have this key present, `Ok(())` is returned and the
```
jonhoo: This is _very_ nitty, but I prefer the active, present voice here (and in general):
> If the map does not contain the key, the key-value pair is inserted and this method returns `Ok`.
>
> If the map does contain the key, the map is left unchanged and this method returns `Err`.

The "map is left unchanged" also covers the `==` case implicitly.

src/map_ref.rs
```diff
@@ -154,6 +154,14 @@ where
         self.map.insert(key, value, &self.guard)
     }
 
+    /// Maps `key` to `value` in this table if, and only if, `key` is not
+    /// already mapped to a value.
+    /// See also [`HashMap::try_insert`].
```
jonhoo: It'd be great if we could leave an empty line after the first sentence of the docs. This makes it a little easier to visually parse.

src/map.rs
```diff
@@ -1348,6 +1348,40 @@ where
         self.put(key, value, false, guard)
     }
 
+    /// Tries to insert a key-value pair into the map.
```
jonhoo: I think I'd like this to read:
> Insert a key-value pair into the map unless the key already exists.

src/map.rs
domenicquirl: Hm, this comes from https://github.com/jonhoo/flurry/commit/23a5fbae32d371da120512778959aa2e4e8de345. I still think `Eq` is clearer... @jonhoo opinion?

src/map_ref.rs
GarkGarcia: No problem.

src/map.rs
GarkGarcia: Done.

src/map.rs
GarkGarcia: > This would still necessitate changes to `put`, right? Because ownership of the new value is transferred to it (so it can create a new node) and we would need to get it back.

Exactly. That's the only way for us to return any information about the inserted value in the success case is to change `put`'s type signature.

GarkGarcia: I personally feel like `Result<(), &V>` is fine.

jonhoo: Yup, it would, which I think is reasonable. We'd change `put` to have an (internal) `enum` return type that indicates the outcome of the `put` call.

src/map.rs
```diff
@@ -1348,15 +1348,13 @@ where
         self.put(key, value, false, guard)
     }
 
-    /// Tries to insert a key-value pair into the map.
+    /// Insert a key-value pair into the map unless the key already exists.
```
jonhoo: Ah, sorry, my bad, this should be "insert**s**"

src/map_ref.rs
```diff
@@ -149,13 +149,15 @@ where
...
+    ///
     /// See also [`HashMap::insert`].
     pub fn insert(&self, key: K, value: V) -> Option<&'_ V> {
         self.map.insert(key, value, &self.guard)
     }
 
-    /// Maps `key` to `value` in this table if, and only if, `key` is not
-    /// already mapped to a value.
+    /// Maps `key` to `value` in this table unless `key` is already mapped to a
+    /// value.
```
jonhoo: Any particular reason why this has a different first sentence from `HashMap::try_insert`?

src/map.rs
jonhoo: I think the `==` comes from the `std` docs, but I agree that `Eq` (or really, `PartialEq`) is more explicit. That said, `==` **is** `PartialEq` in Rust, so :man_shrugging: I'm fine either way, but if @domenicquirl prefers `PartialEq`, then let's go with that. `Eq` I think is misleading.

src/map.rs
jonhoo: I could whip up an updated `put` quickly. Give me a sec.

src/map.rs
domenicquirl: I do (even though I skipped over `PartialEq` as we require `Eq`). I just think it's more clear. Also, we should make it link to the corresponding `std` page.

jonhoo:
Ah, I saw just now that the existing `insert` method has docs similar to the ones I just had you rewrite :p Maybe we should update the `insert` docs to match the style of the new `try_insert` docs while we're at it?

src/map_ref.rs
GarkGarcia: > Any particular reason why this has a different first sentence from `HashMap::try_insert`?

I was trying to base it on the first sentence from `HashMapRef::insert`, but we should probably update them both.

GarkGarcia:
> Ah, I saw just now that the existing `insert` method has docs similar to the ones I just had you rewrite :p Maybe we should update the `insert` docs to match the style of the new `try_insert` docs while we're at it?

Exactly! Gonna work on this.

src/map.rs
GarkGarcia: This isn't an issue anymore for us anymore, since we've altered the wording of the docs and `==` is no longer used. We should probably consider rephrasing `HashMap::insert` documentation though (which still uses `==`).

GarkGarcia: > I do (even though I skipped over `PartialEq` as we require `Eq`). I just think it's more clear. Also, we should make it link to the corresponding `std` page.

Alright. `Eq` it is.

src/map.rs
jonhoo: See https://github.com/jonhoo/flurry/pull/76

src/map.rs
GarkGarcia: Done!

GarkGarcia:
I don't really know why the linting test is failing. I've already runned `cargo fmt`.

src/map.rs
jonhoo: With the reword I still think this ends up going away?

GarkGarcia: > With the reword I still think this ends up going away?

What do you mean? This now only affects the documentation `HashMap::insert`.

jonhoo: Right, but if we're updating the docs of `HashMap::insert` to match the new language of `HashMap::try_insert`, then `Eq` would go away as a consideration there as well.

jonhoo:
#76 has now landed!

src/map.rs
GarkGarcia: I think this has been resolved, hasn't it?

GarkGarcia: > Right, but if we're updating the docs of `HashMap::insert` to match the new language of `HashMap::try_insert`, then `Eq` would go away as a consideration there as well.

I see, I'll have a look at it.

GarkGarcia:
All done I think.

src/map.rs
```diff
@@ -1373,15 +1373,15 @@ where
...
     /// assert_eq!(mref.get(&37), Some(&"a"));
     /// assert_eq!(mref.get(&42), Some(&"c"));
     /// ```
     #[inline]
-    pub fn try_insert<'g>(&'g self, key: K, value: V, guard: &'g Guard) -> Result<(), &'g V> {
+    pub fn try_insert<'g>(&'g self, key: K, value: V, guard: &'g Guard) -> Result<&'g V, &'g V> {
         match self.put(key, value, true, guard) {
-            Some(val) => Err(val),
-            None => Ok(()),
+            PutResult::Replaced { old, .. } | PutResult::Exists { old } => Err(old),
```
jonhoo: When `no_replacement = true`, you should never get `Put::Result::Replaced`, so it'd be good to assert that.

src/map.rs
```diff
@@ -1373,15 +1373,15 @@ where
...
     /// mref.insert(37, "a");
     /// assert_eq!(mref.try_insert(37, "b"), Err(&"a"));
-    /// assert_eq!(mref.try_insert(42, "c"), Ok(()));
+    /// assert_eq!(mref.try_insert(42, "c"), Ok(&"c"));
     /// assert_eq!(mref.get(&37), Some(&"a"));
     /// assert_eq!(mref.get(&42), Some(&"c"));
     /// ```
     #[inline]
-    pub fn try_insert<'g>(&'g self, key: K, value: V, guard: &'g Guard) -> Result<(), &'g V> {
+    pub fn try_insert<'g>(&'g self, key: K, value: V, guard: &'g Guard) -> Result<&'g V, &'g V> {
```
jonhoo: The documentation should probably mention that the `Ok` and `Err` values are.

jonhoo:
I still think we should use the `TryInsertError` struct I proposed earlier, since currently the user has no way to recover the `value` (or `key`) they provided in the case of a failed insert.

jonhoo:
See https://github.com/jonhoo/flurry/pull/77 for a way to recover the value.

GarkGarcia:
> I still think we should use the `TryInsertError` struct I proposed earlier, since currently the user has no way to recover the `value` (or `key`) they provided in the case of a failed insert.

I'm not quite sold on this proposal. Although it does provide some useful functionality which is not provided in our current public API, I seems a bit cumbersome and unergonomic.

Also, since this is a public API, we'd probably have to implement the `?` operator for it, besides some traits for useful conversions.

domenicquirl:
I think functionally the error is similar to a [CompareAndSetError](https://docs.rs/crossbeam-epoch/0.8.2/crossbeam_epoch/struct.CompareAndSetError.html), so that would maybe be another argument for going with a similar implementation as well (CompareAndSetError does contain both old and new value). 

@GarkGarcia what traits exactly would you propose to implement if we went with a custom error?

jonhoo:
@GarkGarcia Ah, so, to be clear, my proposal is:
```rust
pub fn try_insert(..) -> Result<&'g T, TryInsertError<'g, T>> { .. }
```
Which should work just fine with `?`.

GarkGarcia:
> @GarkGarcia Ah, so, to be clear, my proposal is:
> 
> ```rust
> pub fn try_insert(..) -> Result<&'g T, TryInsertError<'g, T>> { .. }
> ```
> 
> Which should work just fine with `?` since we're still using `Result`.

Ohh I see. Seems quite reasonable then.

GarkGarcia:
> I think functionally the error is similar to a [CompareAndSetError](https://docs.rs/crossbeam-epoch/0.8.2/crossbeam_epoch/struct.CompareAndSetError.html), so that would maybe be another argument for going with a similar implementation as well (CompareAndSetError does contain both old and new value).

Fair enought.

> @GarkGarcia what traits exactly would you propose to implement if we went with a custom error?

Mainly `Error`, `Display` and `Debug`, but also `Clone`, `PartialEq` and `Eq` whenever possible.

src/map.rs
```diff
@@ -1388,8 +1388,11 @@ where
     #[inline]
     pub fn try_insert<'g>(&'g self, key: K, value: V, guard: &'g Guard) -> Result<&'g V, &'g V> {
         match self.put(key, value, true, guard) {
-            PutResult::Replaced { old, .. } | PutResult::Exists { old } => Err(old),
+            PutResult::Exists { old, .. } => Err(old),
             PutResult::Inserted { new } => Ok(new),
+            PutResult::Replaced { .. } => unreachable!(
+                "When no_replacement = true, you should never get Put::Result::Replaced"
```
```suggestion
                "no_replacement cannot result in PutResult::Replaced"
```

GarkGarcia:
I don't know why the linting tests are failing, I already run `cargo fmt`.

Also, maybe we should make the fields of `TryInsertError` private. That would make it so that users can only construct an instance of `TryInsertError` through  the `try_insert` methods. We could then allow access to those fields through getters. We would then have to adapt the example of `HashMap::try_insert` though.

domenicquirl:
Can/did you run `cargo fmt --all -- --check` exactly? That's what the CI task checks against.

Regarding the fields, I don't actually know; can you destructure a struct if it's fields are private? I would like the user to be able to do 
```rust
match map.try_insert(&i, value, &guard) {
	Ok(new) => {...},
	Err(TryInsertError { old, not_inserted } ) => {...}
}
```

GarkGarcia:
> Regarding the fields, I don't actually know; can you destructure a struct if it's fields are private? I would like the user to be able to do
> 
> ```rust
> match map.try_insert(&i, value, &guard) {
> 	Ok(new) => {...},
> 	Err(TryInsertError { old, not_inserted } ) => {...}
> }
> ```

Nope, you can't. I think you can deconstruct them inside the `matches!` macro, but I'm not sure and that's pretty limited anyway.

GarkGarcia:
> Can/did you run `cargo fmt --all -- --check` exactly? That's what the CI task checks against.

Just did it:

```
$ cargo fmt --all -- --check
Diff in ~/Documents/flurry/src/map.rs at line 135:

 #[derive(Eq, PartialEq, Clone, Debug)]
 enum PutResult<'a, T> {
-    Inserted {
-        new: &'a T,
-    },
-    Replaced {
-        old: &'a T,
-        new: &'a T,
-    },
-    Exists {
-        old: &'a T,
-        not_inserted: Box<T>,
-    },
+    Inserted { new: &'a T },
+    Replaced { old: &'a T, new: &'a T },
+    Exists { old: &'a T, not_inserted: Box<T> },
 }

 impl<'a, T> PutResult<'a, T> {
```

Dunno why `cargo fmt` hasn't fixed it by itself. Anyway, on my way to fixing it.

domenicquirl:
> Nope, you can't. I think you can deconstruct them inside the `matches!` macro, but I'm not sure and that's pretty limited anyway.

Then I'm against making them private. I think it's not worth for them to be private if we lose the destructuring

GarkGarcia:
> > Nope, you can't. I think you can deconstruct them inside the `matches!` macro, but I'm not sure and that's pretty limited anyway.
> 
> Then I'm against making them private. I think it's not worth for them to be private if we lose the destructuring

Makes sence. @jonhoo what do you think?

src/map.rs
```diff
@@ -168,6 +168,19 @@ impl<'a, T> PutResult<'a, T> {
...
 }
 
+/// The error type for the [`HashMap::try_insert`] method.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct TryInsertError<'a, V> {
+    /// A reference to the old value mapped to the key.
+    pub old: &'a V,
+    /// A reference to the new value, the one passed as an argument to
+    /// [`HashMap::try_insert`].
+    pub not_inserted: &'a V,
```
jonhoo: This should be a `V`, not a `&'a V`.

jonhoo:
I am strongly in favor of both fields being public. We _could_ mark it [`#[non_exhaustive]`](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md), but I don't think even that's necessary.

src/map.rs
jonhoo: (Just dereference the `Box<V>` to move the `V` out of there).

GarkGarcia:
> I am strongly in favor of both fields being public. We _could_ mark it [`#[non_exhaustive]`](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md), but I don't think even that's necessary.

Alright, public it is. Marking things as `#[non_exhaustive]` is good practice.

I guess the only things left to do are:

* Polishing the documentation of `TryInsertError` and `TryInsertResult`.
* Implementing `Display` and `Error` for `TryInsertError`.

src/map.rs
GarkGarcia: Alright.

src/map.rs
```diff
@@ -168,6 +159,19 @@ impl<'a, T> PutResult<'a, T> {
...
+}
+
+/// The return type of [`HashMap::try_insert`].
+pub type TryInsertResult<'a, V> = Result<&'a V, TryInsertError<'a, V>>;
```
jonhoo: I don't think we should have this type alias.

jonhoo:
The biggest downside of `#[non_exhaustive]` is that people now _must_ match on `TryInsertError` with `, ..`, which is a bit of an ergonomic penalty. I guess the question is how likely we think it is that we'll want to change it later. I think it's fairly unlikely..?

src/map.rs
GarkGarcia: I created it out of frustration (of typing `Result<&'a V, TryInsertError<'a, V>>`). What about making it private?

jonhoo: Hmm, that type should only appear in two places though, right? Once in `map_ref.rs` and once in `map.rs`? I don't think adding an extra type to the API is worth the single duplication. We also sadly can't make it private as it would then still appear in the function signatures, but the user wouldn't know what type it actually is.

src/map.rs
GarkGarcia: I can't remember how to unwrap a `Box` hehe. Couldn't find it in the docs either, would you mind helping me out?

Also, has the `box` keyword being stabilized yet? If it is, I think I can just patter match on `box contents_of_the_box`, can't I?

GarkGarcia:
> The biggest downside of `#[non_exhaustive]` is that people now _must_ match on `TryInsertError` with `, ..`, which is a bit of an ergonomic penalty. I guess the question is how likely we think it is that we'll want to change it later. I think it's fairly unlikely..?

Fair enough. Seems unlikely indeed.

src/map.rs
GarkGarcia: Oh, I see. I thought private type aliases were simply ignored by `docs.rs`.

Makes sense, we should expand our API surface unnecessarily. I'll fix this as soon as I can.

jonhoo: Ah, no, sadly not. Type aliases are treated as identical to the original type, but they are not "removed". Whenever they appear, they appear as the alias, not the target type. In fact, the code probably wouldn't even compile since you'd have a private type in a public interface!

src/map.rs
GarkGarcia: > (Just dereference the `Box<V>` to move the `V` out of there).

Has the `box` keyword been stabilized already? I think I could just pattern match on `box inner_contents_of_the_box` if that's the case, couldn't I?

jonhoo: No, I don't think so. Last I heard it also wasn't going to stabilize any time soon. Does `*not_inserted` not work?

src/map.rs
```diff
@@ -166,7 +167,31 @@ pub struct TryInsertError<'a, V> {
...
+}
+
+impl<'a, V> Display for TryInsertError<'a, V>
+where
+    V: Display,
+{
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "The value \"{}\" was inserted into the map, but the value \"{}\" was already associated with that key",
```
jonhoo: This seems a little misleading, since `self.not_inserted` was, well, _not_ inserted.

src/map.rs
GarkGarcia: It works. That's the solution I went for. I was just asking because the `box` thing seemed more ergonomic and, as listener of _Rustacian Station_, I knew you'd probably know the answer to my question ❤️

src/map.rs
GarkGarcia: > This seems a little misleading, since `self.not_inserted` was, well, _not_ inserted.

Indeed! I'm not very good at writing friendly error messages, could you help me out in this one?

Also, I think we should implement `Display` regardless of whether or not `V` implements `Display`. My idea is to provide a generic error message that doesn't take into consideration the `old` and `not_iserted` fields.If the user want's to take those fields into account, he can always access them and create a custom error message.

jonhoo: How about:
> Insert of "{}" failed as key was already present with value "{}"

I think I would probably implement `Display` if `V: Debug`. That seems like a reasonable compromise to me.

src/map.rs
```diff
@@ -168,6 +160,39 @@ impl<'a, T> PutResult<'a, T> {
     }
 }
 
+/// The error type for the [`HashMap::try_insert`] method.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct TryInsertError<'a, V> {
+    /// A reference to the old value mapped to the key.
```
jonhoo: I think I would rename this to `current`, since it is not actually old (as we didn't replace it).

src/map.rs
```diff
@@ -168,6 +160,39 @@ impl<'a, T> PutResult<'a, T> {
...
+    pub old: &'a V,
+    /// A reference to the new value, the one passed as an argument to
```
jonhoo: I would avoid referring to this as "new" in this context, since it's not new as we did not do the replace. Maybe
> The value that [`HashMap::try_insert`] failed to insert.

src/map.rs
```diff
@@ -163,10 +171,9 @@ impl<'a, T> PutResult<'a, T> {
 /// The error type for the [`HashMap::try_insert`] method.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub struct TryInsertError<'a, V> {
-    /// A reference to the old value mapped to the key.
-    pub old: &'a V,
-    /// A reference to the new value, the one passed as an argument to
-    /// [`HashMap::try_insert`].
+    /// A reference to the current value mapped to the key.
+    pub current: &'a V,
+    /// The value that [HashMap::try_insert] failed to insert.
```
```suggestion
    /// The value that [`HashMap::try_insert`] failed to insert.
```

GarkGarcia:
Thanks @jonhoo and @domenicquirl  for the help!

jonhoo:
Thanks for taking the time to implement it, and to follow up on all our nitpicking :sweat_smile:

PR #76, opened 2020-03-23
Make HashMap::put return more info

jonhoo:
Now, `put` will let you know both what it ended up doing (insert,
replace, or nothing if no_replacement is set). This enables it to be
used in more settings. The return value also now carries information
about both the previous and new value for the key, which may be useful
in certain contexts.

Needed for #74

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/76)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -133,6 +133,31 @@ pub struct HashMap<K, V, S = crate::DefaultHashBuilder> {
     build_hasher: S,
 }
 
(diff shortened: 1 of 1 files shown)


GarkGarcia:
Great! As soon as this is merged I'll resume work on #74.

domenicquirl:
Yes, that's exactly what I thought 😊

Issue #20, opened 2020-01-24
Port and add more tests


jonhoo:
There are many Java tests that we do not yet have in the Rust version. You can find most of them in [this file](https://github.com/jonhoo/flurry/blob/master/jsr166/test/ConcurrentHashMapTest.java) from JSR166 and in [this folder](https://hg.openjdk.java.net/jdk/jdk13/file/tip/test/jdk/java/util/concurrent/ConcurrentHashMap) from JDK. Let's port all of them and get that test coverage up! I would particularly love to get some of the concurrent tests ported over!

There are also a good chunk of tests over in [`hashbrown`](https://github.com/rust-lang/hashbrown/tree/master/tests) that would be good to also run against flurry.

jonhoo:
There are even more hashbrown [unit tests here](https://github.com/rust-lang/hashbrown/blob/6f5826ed934a5e6342681c27618bea71f9b0c800/src/map.rs#L2713).

jonhoo:
The C++ concurrent hash table `libcuckoo` has its own set of tests [over here](https://github.com/efficient/libcuckoo/tree/master/tests/unit-tests) that would be great to port!

jonhoo:
`junction` is another C++ concurrent hash table (several actually), and it has a number of tests (and simple benchmarks) [here](https://github.com/preshing/junction/tree/master/samples) that we could probably use as inspiration.

Stupremee:
Some of the java tests also serve as benchmarks. Should we port the benchmarks as well or just the tests?

jonhoo:
Probably both ultimately. Porting them into separate files seems like a good idea. And the benchmark-like tests should be written using criterion so we can run regression tests on them.

Stupremee:
I would like to implement some tests and some benchmarks. Especially a stress test like libcuckoo

jonhoo:
The tests that I know of are the [JDK tests](https://hg.openjdk.java.net/jdk/jdk13/file/tip/test/jdk/java/util/concurrent/ConcurrentHashMap), the [JSR166 tests](https://github.com/jonhoo/flurry/blob/master/jsr166/test/ConcurrentHashMapTest.java), and the [`hashbrown` tests](https://github.com/rust-lang/hashbrown/tree/master/tests). Ideally we'd have pretty much all the tests from all those three.

jonhoo:
As a follow-up to #78, we should also add tests for `HashSetRef`!

PR #78, opened 2020-03-25
Finish fixing #17

GarkGarcia:
This is a follow up from #17.

I messed something up while trying to pull some changes from upstream, so there are 23 commits from my previous contribution in here.

The only relevants commits are the last two.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/78)


jonhoo:
Ah, yes, that's my fault because I ended up doing some squashing. If only the last two matter, run:
```console
$ git remote add upstream https://github.com/jonhoo/flurry.git
$ git fetch upstream
$ git rebase --onto upstream/master HEAD~2
```
That should, in theory, replay just the last two commits onto the current `master` branch. Then you'll have to do:
```console
$ git push --force-with-lease
```
To overwrite your current PR branch with the rebased one.

BTW, if you want to learn more about this sort of magic, I highly recommend [this lecture](https://missing.csail.mit.edu/2020/version-control/).

jonhoo:
@GarkGarcia I just pushed a commit to `master` that re-aligns the methods in `map_ref.rs` with those in `map.rs` to make comparison easier. We should probably do the same here for `set.rs`. I've also added a newline above each "See also ..." line, so that it doesn't get lost in the surrounding text in the docs.

src/set_ref.rs
```diff
@@ -0,0 +1,229 @@
...
+            guard: GuardRef::Ref(guard),
+        }
+    }
+}
+
+impl<T, S> HashSetRef<'_, T, S> {
+    /// An iterator visiting all key-value pairs in arbitrary order.
+    /// The iterator element type is `(&'g K, &'g V)`.
+    /// See also [`HashSet::iter`].
+    pub fn iter(&self) -> Keys<'_, T, ()> {
```
jonhoo: We should do the same re-org for this as I recently did for `map_ref` so that the methods and impl blocks are all in the same order as they are in `set.rs`.

src/set_ref.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+impl<K, V, S> HashSetRef<'_, K, V, S>
+where
+    K: Hash + Eq,
+    S: BuildHasher,
+{
+    /// Tests if `key` is a key in this table.
+    /// See also [`HashSet::contains_key`].
+    pub fn contains<Q>(&self, value: &Q) -> bool
```
twe4ked: I think most (all?) of the `HashSet` docs refer to the items in the set as `values` rather than `keys`. Maybe we should update these to be consistent with that?

GarkGarcia: I haven't paid much attention to the docs yet. I'm trying to make everything compile before that. Still, I'll work on this.

GarkGarcia: Just fixed it in 7bc63b0.

src/set.rs
```diff
@@ -367,6 +371,62 @@ where
...
+    ///
+    /// let set = HashSet::new();
+    ///
+    /// for i in 0..8 {
+    ///     set.pin().insert(i);
+    /// }
+    /// set.pin().retain(|&e| e % 2 == 0);
+    /// assert_eq!(set.pin().len(), 4);
+    /// ```
+    #[inline]
```
jonhoo: This should not be `#[inline]`.

src/set.rs
```diff
@@ -367,6 +371,62 @@ where
...
+    /// ```
+    /// use flurry::HashSet;
+    ///
+    /// let set = HashSet::new();
+    ///
+    /// set.pin().insert("a");
+    /// set.pin().clear();
+    /// assert!(set.pin().is_empty());
+    /// ```
+    #[inline]
```
jonhoo: This should not be `#[inline]`.

src/set.rs
```diff
@@ -367,6 +371,62 @@ where
...
+    pub fn clear(&self, guard: &Guard) {
+        self.map.clear(guard)
+    }
+
+    /// Tries to reserve capacity for at least `additional` more elements to
+    /// be inserted in the `HashSet`. The collection may reserve more space to
+    /// avoid frequent reallocations.
+    #[inline]
```
jonhoo: This should not be `#[inline]`.

src/set.rs
```diff
@@ -367,6 +371,62 @@ where
...
+    /// set.pin().clear();
+    /// assert!(set.pin().is_empty());
+    /// ```
+    #[inline]
+    pub fn clear(&self, guard: &Guard) {
+        self.map.clear(guard)
+    }
+
+    /// Tries to reserve capacity for at least `additional` more elements to
+    /// be inserted in the `HashSet`. The collection may reserve more space to
```
jonhoo: We should add a double-newline here after the first sentence. The second is a note about the implementation.

GarkGarcia: Alright.

GarkGarcia:
We should probably think about implementing [`HashSet::intersection`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.intersection) and [`HashSet::union`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.union) equivalents.

src/set.rs
```diff
@@ -276,6 +276,93 @@ where
...
+    /// assert_eq!(a.is_disjoint(&b, &guard), true);
+    /// b.insert(4, &guard);
+    /// assert_eq!(a.is_disjoint(&b, &guard), true);
+    /// b.insert(1, &guard);
+    /// assert_eq!(a.is_disjoint(&b, &guard), false);
+    ///
+    /// ```
+    pub fn is_disjoint(&self, other: &HashSet<T, S>, guard: &Guard) -> bool {
+        for value in self.iter(guard) {
+            if other.contains(&value, guard) {
```
jonhoo: Hmm, this probably wont' work down the line when we have non-global collectors, since a guard for one map/set won't work on a different one. Probably not worth fixing for now though.

src/set_ref.rs
```diff
@@ -103,22 +103,31 @@ where
     /// Returns `true` if `self` has no elements in common with `other`.
     ///
     /// See also [`HashSet::is_disjoint`].
-    pub fn is_disjoint(&self, other: &HashSet<T, S>) -> bool {
-        self.set.is_disjoint(other, &self.guard)
+    pub fn is_disjoint<Q>(&self, other: &Q) -> bool
+    where
+        Q: AsRef<HashSet<T, S>>,
```
jonhoo: What's the reasoning behind making this change? The `HashSet` API in `std` does not have this generic parameter.

jonhoo:
@GarkGarcia Let's leave those for later so that we do not overload this PR _too_ much.

src/set_ref.rs
GarkGarcia: I made this so that `HashSetRef::is_disjoint` could take either a `&HashSet`or a `&HashSetRef` as an argument.

GarkGarcia:
> @GarkGarcia Let's leave those for later so that we do not overload this PR _too_ much.

Fair enough.

src/set.rs
GarkGarcia: Oh, I see. Is there a workaround? Maybe we should use two different guards?

GarkGarcia: If that's the case, seems like quite an easy fix, I think it's worth it.

src/set_ref.rs
GarkGarcia: Alright. were should I place it then?

src/set_ref.rs
GarkGarcia: If we end up using two guards in `HashSet::is_disjoint`, we'd have to make `HashSetRef::is_disjoint` take a `HashSetRef`. The same goes for `is_subset` and `is_superset`of course.

jonhoo: I'm not sure I follow — why would it need to take `HashSetRef`? The `guard` and `pin` methods should do nicely?

GarkGarcia: > I'm not sure I follow — why would it need to take `HashSetRef`? The `guard` and `pin` methods should do nicely?

Indeed, that would work too. Passing a `HashSetRef` makes more sense to me, as it would allow to user to use different types of guard, but passing a `HashSet`and pinning it works great too.

jonhoo: I think I'm actually coming around to you idea of making it generic over `Q: AsRef<HashMap>`. It does mean that we may end up double-pinning, but that shouldn't be too much of an issue.

GarkGarcia: I went for `HashSetRef`in the last commit, as I felt I wasn't able to express my idea without presenting the code. Keep I mind I can always revert back to the gereric type signature if you guys feel like that's a better fit for our API.

src/set.rs
```diff
@@ -288,18 +288,22 @@ where
...
-    /// b.insert(4, &guard);
-    /// assert_eq!(a.is_disjoint(&b, &guard), true);
-    /// b.insert(1, &guard);
-    /// assert_eq!(a.is_disjoint(&b, &guard), false);
-    ///
-    /// ```
-    pub fn is_disjoint(&self, other: &HashSet<T, S>, guard: &Guard) -> bool {
-        for value in self.iter(guard) {
-            if other.contains(&value, guard) {
+    /// assert_eq!(a.pin().is_disjoint(&b.pin()), true);
```
jonhoo: These can just be `assert!` instead of `assert_eq!` :)

src/set_ref.rs
jonhoo: Actually, how about I just do the re-ordering? Might be faster, since I know what I want :p I _think_ I can just push to your branch.

src/set.rs
GarkGarcia: Sure. I was copying the examples from [`std::collections::HashSet::is_disjoint`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.is_disjoint), which use `assert_eq` instead of `assert`.

jonhoo:
Okay, so, apparently I can't push directly to your branch, but I pushed the added commits to [`pr-17-augment`](https://github.com/jonhoo/flurry/commits/pr-17-augment). You should be able to merge my commits using:
```console
$ git fetch upstream
$ git merge upstream/pr-17-augment
```

jonhoo:
@GarkGarcia In a follow-up PR, it'd be great to also add a test suite for the ref version of set (just like we have one for map)

GarkGarcia:
> @GarkGarcia In a follow-up PR, it'd be great to also add a test suite for the ref version of set (just like we have one for map)

Makes sense. We could probably include the tests in a PR for #20?

Issue #15, opened 2020-01-23
Serialization support


jonhoo:
We should add support for `serde::{Serialize, Deserialize}`. The serde docs on [custom implementations](https://serde.rs/custom-serialization.html) may be helpful, and the [`hashbrown` implementation](https://github.com/rust-lang/hashbrown/blob/b8c34c96924ccfbe79a09d0ffc01d0744a42afcb/src/external_trait_impls/serde.rs) almost certainly will be.

holocenian:
I am very new to Rust. But I would like to help.

jonhoo:
@holocenian Did you end up picking this up? :)

GarkGarcia:
Are you thinking about serializing all four data structures exposed by the crate?

jonhoo:
Yeah, that seems reasonable for `Serialize`. I'd just stick the main impl on the `Ref` types, and then have the non-`Ref` impls just call `self.pin().serialize`. `Deserialize` I think should only be on `HashMap` and `HashSet`.

GarkGarcia:
Makes sense. I'll work on that.

GarkGarcia:
Should we make serialization support option (via a feater)?

jonhoo:
Yes, my inclination would be to make it optional and off by default.

jonhoo:
Fixed in #79 :tada:

PR #79, opened 2020-03-27
Serialization support

GarkGarcia:
This is a follow up to #15.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/79)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ maintenance = { status = "experimental" }
 
 [features]
 sanitize = ['crossbeam-epoch/sanitize']
-serialize = ["serde"]
 
 [dependencies]
 crossbeam-epoch = "0.8.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ serialize = ["serde"]
 crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
-serde = {varsion = "1.0.105", optional = true}
+serde = {version = "1.0.105", optional = true}
 
 [dependencies.ahash]
 version = "0.3.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ default-features = false
 rand = "0.7"
 rayon = "1.3"
 criterion = "0.3"
+serde_json = "1.0.50"
 
 [[bench]]
 name = "flurry_dashmap" 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,12 +21,13 @@ maintenance = { status = "experimental" }
 
 [features]
 sanitize = ['crossbeam-epoch/sanitize']
+serialize = ["serde"]
 
 [dependencies]
 crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
-serde = "1.0.105"
+serde = {varsion = "1.0.105", optional = true}
 
 [dependencies.ahash]
 version = "0.3.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ sanitize = ['crossbeam-epoch/sanitize']
 crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
+serde = "1.0.105"
 
 [dependencies.ahash]
 version = "0.3.2"
--- a/src/map.rs
+++ b/src/map.rs
@@ -2414,7 +2414,7 @@ where
     V: 'static + Deserialize<'de> + Send + Sync + Eq,
     S: Default,
 {
-    type Value = HashMap<K, V, crate::DefaultHashBuilder>;
+    type Value = HashMap<K, V, S>;
 
     // TODO: Create an error message
     fn expecting(&self, _f: &mut Formatter<'_>) -> fmt::Result {
--- a/src/map.rs
+++ b/src/map.rs
@@ -2416,9 +2416,8 @@ where
 {
     type Value = HashMap<K, V, S>;
 
-    // TODO: Create an error message
-    fn expecting(&self, _f: &mut Formatter<'_>) -> fmt::Result {
-        todo!("Create an error message");
+    fn expecting(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "a map could no be deserialized");
     }
 
     fn visit_map<M>(self, mut access: M) -> Result<Self::Value, M::Error>
--- a/src/map.rs
+++ b/src/map.rs
@@ -2,7 +2,7 @@ use crate::iter::*;
 use crate::node::*;
 use crate::raw::*;
 use crossbeam_epoch::{self as epoch, Atomic, Guard, Owned, Shared};
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 use serde::{
     de::{MapAccess, Visitor},
     Deserialize, Deserializer, Serialize, Serializer,
@@ -12,7 +12,7 @@ use std::error::Error;
 use std::fmt::{self, Debug, Display, Formatter};
 use std::hash::{BuildHasher, Hash, Hasher};
 use std::iter::FromIterator;
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 use std::marker::PhantomData;
 use std::sync::{
     atomic::{AtomicIsize, AtomicUsize, Ordering},
@@ -2360,7 +2360,7 @@ where
     }
 }
 
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 impl<K, V, S> Serialize for HashMap<K, V, S>
 where
     K: Serialize,
@@ -2374,7 +2374,7 @@ where
     }
 }
 
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 impl<'de, K, V> Deserialize<'de> for HashMap<K, V, crate::DefaultHashBuilder>
 where
     K: 'static + Deserialize<'de> + Send + Sync + Hash + Clone + Eq,
@@ -2388,13 +2388,13 @@ where
     }
 }
 
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 struct HashMapVisitor<K, V> {
     key_marker: PhantomData<K>,
     value_marker: PhantomData<V>,
 }
 
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 impl<K, V> HashMapVisitor<K, V> {
     pub(crate) fn new() -> Self {
         Self {
@@ -2404,7 +2404,7 @@ impl<K, V> HashMapVisitor<K, V> {
     }
 }
 
-#[cfg(feature = "serialize")]
+#[cfg(feature = "serde")]
 impl<'de, K, V> Visitor<'de> for HashMapVisitor<K, V>
 where
     K: 'static + Deserialize<'de> + Send + Sync + Hash + Clone + Eq,
@@ -2425,9 +2425,7 @@ where
 
         while let Some((key, value)) = access.next_entry()? {
             if let Some(_old_value) = map.insert(key, value, &guard) {
-                unreachable!(
-                    "Serialized map held two values with the same key"
-                );
+                unreachable!("Serialized map held two values with the same key");
             }
         }
 
--- a/src/map.rs
+++ b/src/map.rs
@@ -2,18 +2,11 @@ use crate::iter::*;
 use crate::node::*;
 use crate::raw::*;
 use crossbeam_epoch::{self as epoch, Atomic, Guard, Owned, Shared};
-#[cfg(feature = "serde")]
-use serde::{
-    de::{MapAccess, Visitor},
-    Deserialize, Deserializer, Serialize, Serializer,
-};
 use std::borrow::Borrow;
 use std::error::Error;
 use std::fmt::{self, Debug, Display, Formatter};
 use std::hash::{BuildHasher, Hash, Hasher};
 use std::iter::FromIterator;
-#[cfg(feature = "serde")]
-use std::marker::PhantomData;
 use std::sync::{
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ sanitize = ['crossbeam-epoch/sanitize']
 crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
+serde = "1.0.105"
```
jonhoo: Let's make this optional.

GarkGarcia: Alright.

src/map.rs
```diff
@@ -2367,6 +2371,63 @@ where
...
+
+    fn expecting(&self, _f: &mut Formatter<'_>) -> fmt::Result {
+        todo!("Create an error message");
+    }
+
+    fn visit_map<M>(self, mut access: M) -> Result<Self::Value, M::Error>
+    where
+        M: MapAccess<'de>,
+    {
+        let map = HashMap::with_capacity(access.size_hint().unwrap_or(0));
```
jonhoo: This should probably call `new` if `size_hint().is_none()`, rather than call `with_capacity(0)`. The two are not guaranteed to be the same.

src/map.rs
```diff
@@ -2367,6 +2371,63 @@ where
     }
 }
 
+impl<'de, K, V> Deserialize<'de> for HashMap<K, V, crate::DefaultHashBuilder>
```
jonhoo: I wonder if we might want to also include `S` as generic here `where S: Default`, rather than forcing `DefaultHashBuilder`.

src/map.rs
```diff
@@ -2367,6 +2371,63 @@ where
...
+    where
+        M: MapAccess<'de>,
+    {
+        let map = HashMap::with_capacity(access.size_hint().unwrap_or(0));
+        let guard = map.guard();
+
+        while let Some((key, value)) = access.next_entry()? {
+            if let Some(_old_value) = map.insert(key, value, &guard) {
+                unreachable!(
+                    "`MapAccess::next_entry` should yield two diffirent value associated with the same key"
```
```suggestion
                    "Serialized map held two values with the same key"
```

Cargo.toml
```diff
@@ -21,11 +21,13 @@ maintenance = { status = "experimental" }
 
 [features]
 sanitize = ['crossbeam-epoch/sanitize']
+serialize = ["serde"]
 
 [dependencies]
 crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
+serde = {varsion = "1.0.105", optional = true}
```
jonhoo:
```suggestion
serde = {version = "1.0.105", optional = true}
```

Cargo.toml
```diff
@@ -21,11 +21,13 @@ maintenance = { status = "experimental" }
 
 [features]
 sanitize = ['crossbeam-epoch/sanitize']
+serialize = ["serde"]
```
jonhoo: I think nowadays people just don't explicitly name this feature. By declaring the dependency as optional, people can opt in by naming `serde` as a feature, even though it isn't explicitly listed. See [`hashbrown/Cargo.toml`](https://github.com/rust-lang/hashbrown/blob/master/Cargo.toml).

src/map.rs
```diff
@@ -2,11 +2,18 @@ use crate::iter::*;
 use crate::node::*;
 use crate::raw::*;
 use crossbeam_epoch::{self as epoch, Atomic, Guard, Owned, Shared};
+#[cfg(feature = "serialize")]
```
jonhoo: If we remove the feature and just use `optional` on `serde`, this would become:
```rust
#[cfg(feature = "serde")]
```

src/set.rs
```diff
@@ -600,3 +602,16 @@ where
         }
     }
 }
+
+#[cfg(feature = "serialize")]
+impl<T, S> Serialize for HashSet<T, S>
```
jonhoo: So we're also missing `Deserialize` for `HashSet`, right?

GarkGarcia: Yes, we're still missing that.

Cargo.toml
GarkGarcia: Makes sense. I'll work on this.

src/set.rs
GarkGarcia: I'd like to get some input on the implementation of `Deserialize` before I proceed. Is the `HashMapVisitor` business the right way of going about this?

src/map.rs
GarkGarcia: That would be great. The only issue is that we wouldn't be able to preallocate the map, but that may not be an issue for most use cases (?).

src/map.rs
GarkGarcia: Just fixed it.

src/map.rs
GarkGarcia: Alright. Fixed in the last commit.

src/set.rs
GarkGarcia: I really feel like we're abusing the type system in here, since `HashMapVisitor`does no represent a value. This is apparently how the developers of `serde`expect us to do things, though.

jonhoo: I don't think this is really abusing the type system -- it is helping it! In general, my inclination here is to pretty much exactly whatever [`hashbrown` does for `serde`](https://github.com/rust-lang/hashbrown/blob/master/src/external_trait_impls/serde.rs).

src/map.rs
```diff
@@ -2421,10 +2425,7 @@ where
     where
         M: MapAccess<'de>,
     {
-        let map = match access.size_hint() {
-            Some(n) => HashMap::with_capacity(n),
-            None => HashMap::new(),
-        };
+        let map = S::default();
```
jonhoo: You still probably want to use `HashMap::with_capacity_and_hasher` or `HashMap::with_hasher` here.

src/map.rs
jonhoo: Why not? You can just use `HashMap::with_capacity_and_hasher`?

src/set.rs
GarkGarcia: @jonhoo I'm quite pick about this kind of stuff hehe. I'm quite a fan of the phrase _let data be data_ (and it's counterpart _let procedures be procedures_), so I'm not that comfortable with a type whose only purpose is to encapsulate behaviour.

However, I believe we should go with whatever is the norm. There's no point on trying to fight `serde`'s API and Rust's semantics (which I quite like overall) because of a simple issue like this.

Anyway, enough of my random rambling on semantics. I guess this conversation could be resolved? We've already implemented `Deserialize` for `HashSet`.

src/map.rs
GarkGarcia: I see. You're right! I could use `with_hasher`/`with_capacity_and_hasher` for preallocating the map.

Should we implement `HashMap::with_capacity`, `HashMap::new`and `Default` for `HashMap<K, V, S>` where `S: Default` though? That would make the API a bit more flexible (and I can't see why that wouldn't work).

src/map.rs
GarkGarcia: Indeed! Just fixed this in 9822de1.

GarkGarcia:
I'm not quite sure about how should I fix the _Minimum supported Rust version: 1.36.0_ thing (which is causing CI to block)...

src/map.rs
jonhoo: Ah, no, `with_capacity` and `new` are specifically _not_ generic over `S`, because it leads to the user having to explicitly give `let x: HashMap<K, V> = HashMap::new` everywhere to opt into the default. Having `impl Default` be generic over `S` makes sense to me though.

src/set.rs
jonhoo: I actually don't agree — I think it is extremely useful to be able to leverage the type system to express constraints that are not data-dependent. Especially if it makes illegal states impossible to express. But yes, resolved!

jonhoo:
Well, it fails because of the [use of the `todo!` macro](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1017&view=logs&j=92335eda-b3fc-5c3c-d70a-38a67693cfa7&t=8d1e160f-9194-5495-2c1a-c0e97306df5b), which seems to be easy to just fill in? I think "a map" and "a set" is the style of message that should go in there (again, see `hashbrown`).

GarkGarcia:
> Well, it fails because of the [use of the `todo!` macro](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1017&view=logs&j=92335eda-b3fc-5c3c-d70a-38a67693cfa7&t=8d1e160f-9194-5495-2c1a-c0e97306df5b), which seems to be easy to just fill in? I think "a map" and "a set" is the style of message that should go in there (again, see `hashbrown`).

I see. I didn't know what was causing the block. I'll fix the as soon as I can.

src/map.rs
GarkGarcia: I could work on that - having `impl Default` be generic over `S` - after we're finished with this.

jonhoo:
You can see the build errors by clicking "details" next to the failed test, then "View more details on Azure Pipelines" at the bottom of the shown pane, and then clicking the failing build step in the list on the left.

src/map.rs
```diff
@@ -2416,9 +2416,8 @@ where
 {
     type Value = HashMap<K, V, S>;
 
-    // TODO: Create an error message
-    fn expecting(&self, _f: &mut Formatter<'_>) -> fmt::Result {
-        todo!("Create an error message");
+    fn expecting(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "a map could no be deserialized");
```
jonhoo: I think this should just be "a map". See [`hashbrown`](https://github.com/rust-lang/hashbrown/blob/17572d45dbe31fb1a932c04fe56e9abf777ee4ea/src/external_trait_impls/serde.rs#L62) and the [serde documentation](https://docs.rs/serde/1.0.105/serde/de/trait.Visitor.html#tymethod.expecting):
> The message should complete the sentence "This Visitor expects to receive ...", for example the message could be "an integer between 0 and 64". The message should not be capitalized and should not end with a period.

src/set.rs
```diff
@@ -660,8 +660,8 @@ where
     type Value = HashSet<T, S>;
 
     // TODO: Create an error message
-    fn expecting(&self, _f: &mut Formatter<'_>) -> fmt::Result {
-        todo!("Create an error message");
+    fn expecting(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "a set could not be parsed");
```
jonhoo: Same here — should just be "a set".

src/set.rs
```diff
@@ -660,8 +660,8 @@ where
     type Value = HashSet<T, S>;
 
     // TODO: Create an error message
```
jonhoo: This can also probably go away now.

src/map.rs
```diff
@@ -2417,7 +2417,7 @@ where
     type Value = HashMap<K, V, S>;
 
     fn expecting(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "a map could no be deserialized");
+        write!(f, "a map could no be deserialized")
```
jonhoo: I think this should just be "a map". See [`hashbrown`](https://github.com/rust-lang/hashbrown/blob/17572d45dbe31fb1a932c04fe56e9abf777ee4ea/src/external_trait_impls/serde.rs#L62) and the [serde documentation](https://docs.rs/serde/1.0.105/serde/de/trait.Visitor.html#tymethod.expecting):
> The message should complete the sentence "This Visitor expects to receive ...", for example the message could be "an integer between 0 and 64". The message should not be capitalized and should not end with a period.

src/set.rs
```diff
@@ -661,7 +661,7 @@ where
...
     fn expecting(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "a set could not be parsed");
+        write!(f, "a set could not be parsed")
```
jonhoo: Same here — should just be "a set".

src/map.rs
GarkGarcia: Ohh I see. Just fixed it 😁️

jonhoo:
Okay, one last question before we merge this: do we want to separate the impls of external traits (like `serde`) into a separate file the way hashbrown does? Might help avoid polluting `map.rs` and `set.rs` over time?

jonhoo:
Oh, and we should probably add some rudimentary tests here. A round-trip test for each data structure should be sufficient (that is, that `deserialize(x.serialize()) == x`).

GarkGarcia:
> Okay, one last question before we merge this: do we want to separate the impls of external traits (like `serde`) into a separate file the way hashbrown does? Might help avoid polluting `map.rs` and `set.rs` over time?

Makes sense to me. I'll work on that.

GarkGarcia:
> Oh, and we should probably add some rudimentary tests here. A round-trip test for each data structure should be sufficient (that is, that `deserialize(x.serialize()) == x`).

That would be nice. Is there a way I can import a crate only in the context of test though? I'd have to use something like `serde_json`, but we should have it as a dependency (only a dev-dependency).

jonhoo:
I'm not sure I follow? Just add it as a `dev-dependency`, and you should be fine.

GarkGarcia:
> I'm not sure I follow? Just add it as a `dev-dependency`, and you should be fine.

Yes, I just don't know how to do that hehe.

GarkGarcia:
I've pushed a commit to GarkGarcia:master, but I doesn't show up in here for some reason...

jonhoo:
GitHub is having issues :)

To add a test dependency, just list it under here:
https://github.com/jonhoo/flurry/blob/d1f67c2f40cca3290af2fc8cbcc033e6d0579729/Cargo.toml#L34

GarkGarcia:
> To add a test dependency, just list it under here:
> 
> https://github.com/jonhoo/flurry/blob/d1f67c2f40cca3290af2fc8cbcc033e6d0579729/Cargo.toml#L34

Thanks!

> GitHub is having issues :)

Ohh I see. I found it strange that my commit was rejected a couple of times due to _Internal Server Issues_.

GarkGarcia:
All done (?)

jonhoo:
Nice, thank you!

PR #82, opened 2020-04-10
Stop ignoring miri leaks

jonhoo:
Now that https://github.com/rust-lang/miri/issues/940 has been fixed,
and has landed in nightly
(https://github.com/rust-lang/rust/pull/70897), we should be able to run
miri with leak check enabled again!

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/82)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -3538,6 +3538,7 @@ mod tree_bins {
     }
     #[test]
     #[should_panic]
+    #[cfg(not(miri))]
     fn disallow_evil() {
         let map: HashMap<_, _> = HashMap::default();
         map.insert(42, String::from("hello"), &crossbeam_epoch::pin());
--- a/src/map.rs
+++ b/src/map.rs
@@ -3383,6 +3383,7 @@ mod tree_bins {
     }
 
     #[test]
+    #[cfg(not(miri))]
     fn concurrent_tree_bin() {
         let map = HashMap::<usize, usize, _>::with_hasher(ZeroHashBuilder);
         // first, ensure that we have a tree bin
--- a/src/map.rs
+++ b/src/map.rs
@@ -7,10 +7,7 @@ use std::error::Error;
 use std::fmt::{self, Debug, Display, Formatter};
 use std::hash::{BuildHasher, Hash, Hasher};
 use std::iter::FromIterator;
-use std::sync::{
-    atomic::{AtomicIsize, AtomicUsize, Ordering},
-    Once,
-};
+use std::sync::atomic::{AtomicIsize, Ordering};
 
 const ISIZE_BITS: usize = core::mem::size_of::<isize>() * 8;
 
@@ -62,8 +59,10 @@ const MAX_RESIZERS: isize = (1 << (ISIZE_BITS - RESIZE_STAMP_BITS)) - 1;
 /// The bit shift for recording size stamp in `size_ctl`.
 const RESIZE_STAMP_SHIFT: usize = ISIZE_BITS - RESIZE_STAMP_BITS;
 
-static NCPU_INITIALIZER: Once = Once::new();
-static NCPU: AtomicUsize = AtomicUsize::new(0);
+#[cfg(not(miri))]
+static NCPU_INITIALIZER: std::sync::Once = std::sync::Once::new();
+#[cfg(not(miri))]
+static NCPU: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
 
 macro_rules! load_factor {
     ($n: expr) => {
--- a/src/map.rs
+++ b/src/map.rs
@@ -3383,7 +3383,7 @@ mod tree_bins {
     }
 
     #[test]
-    #[cfg(not(miri))]
+    #[cfg_attr(miri, ignore)]
     fn concurrent_tree_bin() {
         let map = HashMap::<usize, usize, _>::with_hasher(ZeroHashBuilder);
         // first, ensure that we have a tree bin
@@ -3538,7 +3538,7 @@ mod tree_bins {
     }
     #[test]
     #[should_panic]
-    #[cfg(not(miri))]
+    #[cfg_attr(miri, ignore)]
     fn disallow_evil() {
         let map: HashMap<_, _> = HashMap::default();
         map.insert(42, String::from("hello"), &crossbeam_epoch::pin());
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -58,7 +58,7 @@ jobs:
      - script: cargo miri setup
        displayName: cargo miri setup
      - script: cargo miri test
-       displayName: cargo miri test
+       displayName: cargo miri test -- -Zmiri-ignore-leaks
  - job: asan
    dependsOn: deny
    displayName: "Run address sanitizer on test suite"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -57,11 +57,7 @@ jobs:
            - miri
      - script: cargo miri setup
        displayName: cargo miri setup
-     # ignore leaks due to
-     # https://github.com/crossbeam-rs/crossbeam/issues/464
-     # which is
-     # https://github.com/rust-lang/miri/issues/940
-     - script: cargo miri -Zmiri-ignore-leaks test
+     - script: cargo miri test
        displayName: cargo miri test
  - job: asan
    dependsOn: deny
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -57,8 +57,8 @@ jobs:
            - miri
      - script: cargo miri setup
        displayName: cargo miri setup
-     - script: cargo miri test
-       displayName: cargo miri test -- -Zmiri-ignore-leaks
+     - script: cargo miri test -- -Zmiri-ignore-leaks
+       displayName: cargo miri test
  - job: asan
    dependsOn: deny
    displayName: "Run address sanitizer on test suite"
--- a/tests/basic.rs
+++ b/tests/basic.rs
@@ -206,7 +206,7 @@ fn compute_if_present_remove() {
 }
 
 #[test]
-#[cfg(not(miri))]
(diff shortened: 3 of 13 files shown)


jonhoo:
@domenicquirl err, got a spurious `SIGSEGV` on the test suite here :/
https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1068&view=logs&j=d84406a3-b0d8-5688-9074-4fc79d957ac9&t=d3f3159f-94fd-553d-ea32-0524cc49969a
Can you try running the tests in a loop with your latest tree stuff and see if it will occasionally crash?

domenicquirl:
Oh noes! The location of the error in the Linux build you linked looks indeed like it could be a tree test. However, I see you have disabled the concurrent one if miri is active, so it would have to be a single-threaded one. The miri task fails in the `disallow_evil` test (which seems to have slipped into the `treebin` test mod, sorry 'bout that), though that error says it is "experimental", not sure what that's about.

Either way, I've just had `cargo test` run for about 15 minutes in a loop on `master`, but could not reproduce the error (Windows). Have you tried on a Linux machine, or maybe with nightly? Have you seen it occur on other platforms/toolchains while working on this?

jonhoo:
@domenicquirl The error actually occurred in the normal test suite, not the miri one, so could still be the concurrent tree test. The failure was on nightly Linux, so I'll try running it in a loop on my computer too.

Looks like miri still thinks [there is a leak](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1071&view=logs&j=51a00aa1-06f5-5fab-feeb-339bdc13aef3&t=79951b30-7f39-53ec-11c6-4eb34331b41a&l=121). I think until `crossbeam` runs their tests through `miri`'s leak check successfully, we shouldn't rely on it either.

jonhoo:
@domenicquirl managed to reproduce the failed test here: https://github.com/jonhoo/flurry/issues/83

domenicquirl:
Yeah, it looks like every defer inside crossbeam gets tagged as a leak. I guess at least miri did not detect any additional leaks (where we forget to `defer_destroy` something)? But we should probably just re-purpose this to exclude concurrent tests from miri (which is also kind of sad given that they are a huge part of flurry and our tests).

jonhoo:
Hmm, it looks like currently we get stuck on https://github.com/tkaitchuck/aHash/issues/32, so we need to wait for a resolution to that. I wonder how miri stopped working so completely without us noticing?

domenicquirl:
It did run successfully in #79, which is not that old...

jonhoo:
I think it may actually have failed to run, maybe due to some incorrect arguments? Because its [log output](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1028&view=logs&jobId=51a00aa1-06f5-5fab-feeb-339bdc13aef3&j=51a00aa1-06f5-5fab-feeb-339bdc13aef3&t=79951b30-7f39-53ec-11c6-4eb34331b41a) is empty...

src/map.rs
```diff
@@ -3357,6 +3357,7 @@ mod tree_bins {
     }
 
     #[test]
+    #[cfg(not(miri))]
```
RalfJung, line 3360: FWIW, the recommended way to ignore tests on Miri is
```rust
#[cfg_attr(miri, ignore)]
```

jonhoo:
@RalfJung Something weird is going on here now. If I try to run the miri tests, it seems to get stuck somewhere when running the tests. Unfortunately, I cannot pass `--nocapture` to make it print the name of the next test it's going to run, since miri does not seem to recognize that argument. Using a mix of `-v` and the regular test order, I _believe_ the issue is the `hasher` tests in `tests/basic_ref.rs`, but I cannot get miri to run just that test (or even just `tests/basic_ref.rs`). It seems to always insist to at least run the `--lib` tests?

```console
$ cargo +nightly miri test --test basic_ref
```

runs the `--lib` and `basic_ref` tests seemingly _concurrently_, making it very hard to recognize where the hanging test is going. Or indeed to get any useful debug output. Trying to run just the hasher tests gives:

```console
$ cargo +nightly miri test hasher
error: Found argument 'hasher' which wasn't expected, or isn't valid in this context

USAGE:
    cargo check --lib --profile <PROFILE-NAME>

For more information try --help
```

Is there a good other way for me to figure out _which_ test is hanging?

RalfJung:
Maybe the arguments get sent to the wrong program? Quoting from the docs:

> You can pass arguments to Miri after the first --, and pass arguments to the interpreted program or test suite after the second --.

I never used `--lib` or so, but filtering works fine for me with `cargo miri test -- -- filter`.

RalfJung:
Oh also when concurrency is involved and things get stuck, maybe you are running into https://github.com/rust-lang/miri/issues/1388.

RalfJung:
> Unfortunately, I cannot pass --nocapture to make it print the name of the next test it's going to run, since miri does not seem to recognize that argument. 

This is a test suite argument, right? Miri doesn't do anything there, so probably that is also just a case of not enough `--`.

One issue is that unfortunately cargo line-buffers Miri's output, so while a test runs we cannot yet see its name. I don't know how to make cargo not do that -- cargo *thinks* it is running rustc when it actually runs Miri, so line-buffering makes sense for it. I don't know any other way to hook into cargo than overwriting the `RUSTC` env var.

One thing I sometimes do is to run `cargo miri test -v`, which will print the invocation of the underlying Miri driver:
```
"/home/r/.cargo/bin/miri" "-Zalways-encode-mir" "-Zmir-emit-retag" "-Zmir-opt-level=0" "--cfg=miri" "-Cdebug-assertions=on" "--crate-name" "test" "--edition=2018" "tests/test.rs" "--error-format=json" "--json=diagnostic-rendered-ansi" "--emit=dep-info,metadata" "-C" "debuginfo=2" "--test" "-C" "metadata=63848be0a524aed1" "-C" "extra-filename=-63848be0a524aed1" "--out-dir" "/home/r/src/rust/miri/test-cargo-miri/target/debug/deps" "-C" "incremental=/home/r/src/rust/miri/test-cargo-miri/target/debug/incremental" "-L" "dependency=/home/r/src/rust/miri/test-cargo-miri/target/debug/deps" "--extern" "byteorder=/home/r/src/rust/miri/test-cargo-miri/target/debug/deps/libbyteorder-76227ec053bbc4f4.rmeta" "--extern" "num_cpus=/home/r/src/rust/miri/test-cargo-miri/target/debug/deps/libnum_cpus-ced524b066799ccf.rmeta" "--extern" "rand=/home/r/src/rust/miri/test-cargo-miri/target/debug/deps/librand-a1a8ab16d61e5ca4.rmeta" "--sysroot" "/home/r/.cache/miri/HOST"
```
Then I copy-paste that into the shell to run Miri without cargo buffering it. (No env vars should be needed.)

jonhoo:
Hmm, interesting:
```console
$ cargo +nightly miri test -- --test basic_ref
    Checking flurry v0.3.0 (/home/jon/dev/streams/flurry)
error: Option 'test' given more than once

error: could not compile `flurry`.
$ cargo +nightly miri test -- hasher
    Checking flurry v0.3.0 (/home/jon/dev/streams/flurry)
error: multiple input filenames provided (first two filenames are `src/lib.rs` and `hasher`)

error: could not compile `flurry`.

To learn more, run the command again with --verbose.
```
I think there's definitely something funky going on there.

I also tried
```console
$ cargo +nightly miri test -- -- --test basic_ref
```

But that just ran _all_ the test targets (but no tests within each target), making it seem like it was treated as a filtering argument? This seems to be supported by `-- -- hasher` only running the hasher tests. It seems like miri decides what targets to build/run independently of any arguments, and then just blindly passes all the trailing arguments to every invocation?

jonhoo:
Your point about how to avoid buffering seems to have worked. I wonder how this normally works when you run `cargo test -- --nocapture`, because that seems to print partial lines?

jonhoo:
@RalfJung the hanging tests were (luckily) just due to high iteration counts, which were too slow for miri to get through in a reasonable amount of time. See https://github.com/jonhoo/flurry/pull/82/commits/d0f080b813a496dcc287348e786207bc0206d5da

RalfJung:
`cargo +nightly miri test -- --test basic_ref` is passing `--test` to Miri, so this is the rustc `--test` flag you are setting -- which as the message says already got said by cargo, this is not what you want.

I don't know which binary you want to pass `--test` to, cargo, Miri or the test runner itself -- depending on what you want, you need to put it at the right spot between the `--`. I have never seen that parameter for test runs before.

`cargo test -- --nocapture` passes `--nocapture` to the test runner, so the Miri equivalent is `cargo test -- -- --nocapture`.

> I wonder how this normally works when you run cargo test -- --nocapture, because that seems to print partial lines?

Then cargo knows it's running a user binary and doesn't do buffering. But with `cargo miri test`, under the hood the actual command that's being run is `cargo build <flags>`, and then when cargo thinks it is building the binary, really Miri just runs it. But then cargo still does buffering.

jonhoo:
`--test` goes to `cargo build`, but I couldn't find any place to put it where it did the right thing. For context, `--test foo` will build and run _only_ `tests/foo.rs`. `--test` cannot be passed to `rustc`, since `rustc` is just told to compile a particular file. It cannot be passed to the test binary, since at that point it's too late — that test binary has already been compiled.

As for `--nocapture`, I was more musing as to how it worked to see if miri could do something similar. `--nocapture` to a miri test binary probably won't have much of an effect since, as you say, the output is buffered by cargo.

RalfJung:
So `cargo test --test foo` is a thing? Yeah Miri doesn't support that, I think that would be https://github.com/rust-lang/miri/issues/739.

> --nocapture to a miri test binary probably won't have much of an effect since, as you say, the output is buffered by cargo.

Well it still affects whether the test runner itself captures and buffers stdout of the test methods (and suppresses it AFAIK for passing tests).

jonhoo:
Yup, it's a thing, and a fairly handy one at that! I'll move the discussion to that issue, as it seems exactly appropriate.

PR #87, opened 2020-04-13
no longer panic upon seeing a moved or treeified bin in treeify_bin

domenicquirl:
Removes the erroneous assumption that attempts to treeify a bin should always be successful and find a regular (linear) bin when loading the bin from the table.

Instead, it is also possible to find a `Moved` entry if the bin was moved in the meantime, or a `Tree` if another thread already treeified the bin before the first thread got to that point.

Loading a `TreeNode` remains an error, since these shouldn't be the head of a bin.

Closes #83. Refer to this issue or the comment introduced for the above cases with this commit for further explanation.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/87)

PR #88, opened 2020-04-13
store element count as isize and clamp len to >= 0

### LINKED ISSUE #86: map::tree_bins::concurrent_tree_bin: attempt to subtract with overflow
Hit _two_ of these at the same time. This is _after_ #85.
```
test map::tree_bins::concurrent_tree_bin ...
thread '<unnamed>' panicked at 'attempt to subtract with overflow', src/map.rs:1163:17
stack backtrace:
...
  13: core::panicking::panic
             at src/libcore/panicking.rs:54
  14: flurry::map::HashMap<K,V,S>::add_count
             at src/map.rs:1163
  15: flurry::map::HashMap<K,V,S>::replace_node
             at src/map.rs:2626
  16: flurry::map::HashMap<K,V,S>::remove
             at src/map.rs:2366
  17: flurry::map::tree_bins::concurrent_tree_bin::{{closure}}
             at src/map.rs:3429
```
```
thread '<unnamed>' panicked at 'attempt to add with overflow', src/map.rs:1159:17
...
  13: core::panicking::panic
             at src/libcore/panicking.rs:54
  14: flurry::map::HashMap<K,V,S>::add_count
             at src/map.rs:1159
  15: flurry::map::HashMap<K,V,S>::put
             at src/map.rs:1970
  16: flurry::map::HashMap<K,V,S>::insert
             at src/map.rs:1625
  17: flurry::map::tree_bins::concurrent_tree_bin::{{closure}}
             at src/map.rs:3419
```

domenicquirl:
Changes the map's element count to be stored as a signed integer, to avoid panics if the count is updated by for example two removing and one inserting thread out of order.

In order to not report a negative size to users, `len` is adjusted to return `0` if the actual count is negative. This is translated from [the Java implementation](https://github.com/jonhoo/flurry/blob/master/jsr166/src/ConcurrentHashMap.java#L880), but without the upper bounds check, since for us the storage data type (`isize`) cannot be larger than the maximal value of the return type (`usize`).

Note that this means that technically the same race condition still exists at the upper bound, however the map's `MAXIMUM_CAPACITY` is way below `isize::MAX`.

Closes #86.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/88)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -92,7 +92,7 @@ pub struct HashMap<K, V, S = crate::DefaultHashBuilder> {
     /// The next table index (plus one) to split while resizing.
     transfer_index: AtomicIsize,
 
-    count: AtomicUsize,
+    count: AtomicIsize,
 
     /// Table initialization and resizing control.  When negative, the
     /// table is being initialized or resized: -1 for initialization,
@@ -298,7 +298,7 @@ impl<K, V, S> HashMap<K, V, S> {
             table: Atomic::null(),
             next_table: Atomic::null(),
             transfer_index: AtomicIsize::new(0),
-            count: AtomicUsize::new(0),
+            count: AtomicIsize::new(0),
             size_ctl: AtomicIsize::new(0),
             build_hasher: hash_builder,
             collector: epoch::default_collector().clone(),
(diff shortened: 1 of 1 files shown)


domenicquirl:
@jonhoo would you mind giving these changes another test run? I ran our tests on a branch with both this and #87 applied for a good half an hour and got no hits, but I also wasn't able to reproduce the errors this is supposed to fix before, so, yeah.

jonhoo:
Yup — I'll merge them both, and then run the test suite for a while. Thanks!

Issue #14, opened 2020-01-23
Support rayon parallel iterators


jonhoo:
We should implement [`rayon` parallel iterators]. The notes and code in [`iter/plumbing`] may be helpful.

  [`rayon` parallel iterators]: https://docs.rs/rayon/1/rayon/iter/index.html
  [`iter/plumbing`]: https://github.com/rayon-rs/rayon/tree/master/src/iter/plumbing

cuviper:
I'm happy to answer rayon questions on this.

jonhoo:
@cuviper Since you're offering, one thing I was actually wondering was whether we can somehow take advantage of the fact that the map supports fully concurrent access. Since non-concurrent maps like `hashbrown` are also able to implement the parallel traits, it wasn't immediately clear to me whether how you get a win from `flurry` in `rayon` even though it really _feels_ like it should be possible.

jonhoo:
I suspect maybe the answer has to do with some of the traits in `iter/plumbing`, but not sure?

cuviper:
Concurrent `insert` will make a naive `ParallelExtend` and `FromParallelIterator` trivial, just something like `par_iter.for_each(|(key, value)| { map.insert(key, value); })`. Maybe there's a more advanced approach that could improve performance, like folding into separate bins and then reducing into the final map, but I don't know your data structure enough to evaluate that.

Parallel iterators are a bit harder -- you need a strategy for splitting the map into separate "slices"/"views" of some sort. The hashbrown implementation should be a good reference if you look at how they use `RawIterRange::split` internally.

jonhoo:
Ah, I see, `hashbrown` is forced for first [reduce](https://github.com/rust-lang/hashbrown/blob/b8c34c96924ccfbe79a09d0ffc01d0744a42afcb/src/external_trait_impls/rayon/helpers.rs#L7) and then use one core to build the map, whereas we don't have to do that. Neat!

Stupremee:
I would like to give this a try.

jonhoo:
All yours!

Stupremee:
I need some help. Whats the best way to split a Table into two halves?

jonhoo:
The best way is probably to just split the list of bins in two: the "high" bins and the "low" bins.

Stupremee:
I have no idea how to approach this. I think it's better if I let someone else do it.

PR #89, opened 2020-05-26
Initial `rayon` support: ParallelExtend + FromParallelIterator

Others:
This was a straightforward step towards #14.

I'm interested in trying to get some parallel iterator work going, but I thought I'd start with some small stuff :)

Points that need addressing:
- `par_extend_sync` is a crap name, so we likely need to think of something better!
- we create a `guard` for every insert operation which sucks, and I wonder if we can do better
(this is especially relevant since I believe there's going to be a similar problem with parallel iterator support)

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/89)


jonhoo:
@cuviper Since you have the most rayon experience (I believe), would you mind taking a look at this?

Others:
@jonhoo No problem. Do you have any suggestions for getting my tests running on CI? They’re behind a feature flag so I don’t think they’re being picked up.

jonhoo:
By default the normal test suite should be run with `--all-features` I believe. If you want them to also run with miri/ASAN/LSAN, you'll want to modify these lines:
https://github.com/jonhoo/flurry/blob/448d9f6fde3147edda9443c261e2a8c563a8a0c9/azure-pipelines.yml#L60
https://github.com/jonhoo/flurry/blob/448d9f6fde3147edda9443c261e2a8c563a8a0c9/azure-pipelines.yml#L76
https://github.com/jonhoo/flurry/blob/448d9f6fde3147edda9443c261e2a8c563a8a0c9/azure-pipelines.yml#L94
To also include your feature (or just make them `--all-features` — that seems reasonable too).

Others:
Ah, I guess they're running but not picked up by `codecov`?
I'm assuming MIRI isn't going to work, as the tests use threading. I'll add `--all-features` to the sanitizer runs. 

I've been thinking a bit about having to create a guard for every element we want to add. Creating a guard is a bit expensive (it definitely has a performance impact on my project using `flurry`) and might be dragging down performance. We could consider using [thread-local](https://github.com/Amanieu/thread_local-rs) to optimize into just creating one guard per worker thread. Whoever reviews this might want to consider whether that optimization is worth it.

jonhoo:
Ahhh, yeah, that's awkward. Looks like the template we're using [doesn't pass `--all-features` to tarpaulin](https://github.com/crate-ci/azure-pipelines/blob/38f20d787869847076c71943e80063520b532da9/coverage.yml#L51). Easiest way to fix that is probably to just inline the code to run it. We won't need the templating stuff or any of the token detection stuff, so only the last job should be relevant.

So, the guard stuff is actually pretty thorny :sweat_smile: We've discussed it a lot over in https://github.com/jonhoo/bustle/issues/2#issuecomment-628673831 (and the following comments) which you may want to take a look at. I think we should probably not make a change to how guards work in _this_ PR. Definitely agree that we should be able to do better than creating a guard for every _access_ in rayon though. @cuviper does `rayon` have a way to do some "prep" on each thread that is running the parallel iterator?

Others:
Yeah my initial idea to use a thread local on the stack doesn't work, as `Guard` is not Send. (Which makes sense, since it pins a thread.)

src/rayon_impls.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::hash::{BuildHasher, Hash};
+
+impl<K, V, S> ParallelExtend<(K, V)> for HashMap<K, V, S>
+where
+    K: Clone + Hash + Ord + Send + Sync + 'static,
+    V: Send + Sync + 'static,
+    S: BuildHasher + Sync,
+{
+    // This is of limited use due to the `&mut self` parameter. See `par_extend_sync`
+    fn par_extend<I>(&mut self, par_iter: I)
```
cuviper, line 12: It's still of use in that it can be filled concurrently for _this_ extend, even if it requires exclusive access up front.

Another possibility is to `impl ParallelExtend for &HashMap`, so the `&mut self` is really `&mut &HashMap`. You could do that in addition to this base implementation, replacing `par_extend_sync`, although I'm not sure off-hand how well auto-ref/deref will let that play out.

Still, you do want that base implementation so things like `partition` and `unzip` can work.

src/rayon_impls.rs
```diff
@@ -0,0 +1,119 @@
...
+    // FIXME: Terrible name, just didn't want to shadow the rayon name
+    fn par_extend_sync<I>(&self, par_iter: I)
+    where
+        I: IntoParallelIterator<Item = (K, V)>,
+    {
+        par_iter.into_par_iter().for_each(|(k, v)| {
+            // Unfortunate that we need to create a guard for each insert operation
+            // Ideally we'd create a guard for each `rayon` worker thread instead
+            // Perhaps this could be done with a thread local?
+            let guard = self.guard();
```
cuviper, line 35: I think [`for_each_init`](https://docs.rs/rayon/1.3.0/rayon/iter/trait.ParallelIterator.html#method.for_each_init) would be perfect for creating that guard. It's per job, not exactly per thread, which is different depending on how adaptive splitting and work stealing plays out, but close enough most of the time.

src/rayon_impls.rs
```diff
@@ -0,0 +1,119 @@
...
+            self.insert(k, v, &guard);
+        });
+    }
+}
+
+impl<K, V> FromParallelIterator<(K, V)> for HashMap<K, V, crate::DefaultHashBuilder>
```
cuviper, line 41: Why isn't this parameterized on `S`? If you compare to the standard `HashMap`, that uses `S: BuildHasher + Default`, so you can just create a `HashMap::default()` and then `insert`/`extend` everything.

src/rayon_impls.rs
Others: Yeah, I more meant "this is not something the library user will usually want to use directly." I think almost everyone who uses `flurry` wraps it in an `Arc` immediately. You're right it is important to have it of course. (I removed the comment.)

Nice call with the `impl` for `&HashMap`! I feel smarter having learned this trick

src/rayon_impls.rs
```diff
@@ -0,0 +1,117 @@
+use crate::HashMap;
+use rayon::iter::{FromParallelIterator, IntoParallelIterator, ParallelExtend, ParallelIterator};
+use std::hash::{BuildHasher, Hash};
+
+impl<K, V, S> ParallelExtend<(K, V)> for HashMap<K, V, S>
```
jonhoo, line 5: Could we implement this for `HashMapRef` as well?

cuviper: Not really, since its `Guard` is stuck on one thread.

I guess you could hack it by ignoring the ref's own guard and extending the map inside.

jonhoo: Yeah, that was basically what I was thinking. Not sure how much that _really_ buys us, but if we're trying to encourage users to use `HashMapRef` over dealing with guards themselves, it might be a nice ergonomic improvement.

cuviper: FWIW, `HashMapRef` doesn't even have plain `Extend` yet. :wink: 

The `HashMap` implementations don't make you deal with guards anyway, since the trait methods don't really make that possible, but it seems fine to keep API parity and add that to `HashMapRef` too.

While we're talking parity, there's only `Extend<(K, V)> for &HashMap`, not `for HashMap`, but this PR has `ParallelExtend` for both. But there is also `Extend<(&K, &V)> for &HashMap` with `Copy` types, which this PR doesn't have.

Then multiply all of this for sets...

jonhoo: Bah... I'm honestly not sure how best to deal with that. I'm inclined to remove the `where Copy` impl. The impls for `&HashMap` I think make sense for us (arguably more so than "regular" `Extend`). I think we could go as far as to remove the `impl for HashMap` entirely, and _only_ implement for `&`.

cuviper: Iterator methods like `partition` and `unzip` need `Extend` for the collection itself. If you only implement for `&`, then it's basically only useful for direct `extend(...)` calls.

That matters less for `HashMapRef`, which is already a reference of sorts. I would just do that one directly, not `for &HashMapRef`, but it can't be a target of `unzip`/etc. anyway.

`Copy` impls are debatable, but they're at least consistent with `std` collections.

jonhoo: Oh, interesting, I didn't even know that. I suppose they all sort of need to stay then (well, `Copy` I think is still unclear, given we now have `Iterator::copied`). I agree with you that impls for `&HashMapRef` are probably unnecessary.

cuviper: Well, `copied()` doesn't work for `Item=(&K, &V)` since that tuple itself is not a reference.

jonhoo: The heat in Boston is clearly getting to me. And that explains why the standard library `HashMap` has an implementation for `Copy` (and why we probably should too).

Others:
Just pinging that this is still on my radar, will hopefully get to this in the later half of the week :)

Others:
Okay I think this hits a decent subset of the impls we want, plus mostly okay tests. Might be suffering a bit of PR creep, so I'd rather not make it any bigger.

jonhoo:
This looks good to me — thanks!

cheako:
One of Rust's features is the naming convention, it seems clumsy and random(like a blaster).  That has both pro(s) and con(s), like not knowing what something is.  I think the OP(@Others @jonhoo) here should be extended with a short description of what is the goal this PR is trying to accomplish...  With an intention on explaining it to someone who will then follow newly added links to rayon site(s) for deeper explanation and further reading I.E. someone who thinks rayon has more todo with light than as a Rust crate.

```
Walk entries on multiple threads using convenience [rayon](https://docs.rs/rayon) wrapper [trait](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html).  Also consume same trait for [insertion](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelExtend.html) on multiple threads.
```

jonhoo:
I'm not exactly sure what you're asking for here? If you'd like to see documentation on a particular method be improved, then I'd be happy to take a PR with that improvement!

cheako:
The goal was unclear without prior knowledge of rayon or whatever.  Without links to rayon that leads to googling or looking to see if there is a docs page.

jonhoo:
The goal of what exactly? Of the PR? Of the commit? Or of the interface in the code? I don't generally expect contributors to document their PRs/commits directly, since it's assumed that those will be seen by developers of the library, not of its users.

PR #91, opened 2020-08-28
Early return in the slow path of `put` if key already exists and `no_replacement` is true

### LINKED ISSUE #90: HashMap enters unreachable code in try_insert
Code to reproduce: 
```
use flurry::HashMap;
use rand::{thread_rng, Rng};
use flurry::epoch::pin;

fn main() {
    let mut rng = thread_rng();
    let map = HashMap::new();
    let g = pin();
    for _ in 0..10000 {
        let el = rng.gen_range(0, 1000);
        let _ = map.try_insert(el, el, &g);
    }
}
```
Output:
```
thread 'main' panicked at 'internal error: entered unreachable code: no_replacement cannot result in PutResult::Replaced', <::std::macros::panic macros>:5:6
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1052
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:224
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:472
  11: rust_begin_unwind
             at src/libstd/panicking.rs:380
  12: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:334
  13: flurry::map::HashMap<K,V,S>::try_insert
             at ./<::std::macros::panic macros>:5
  14: flurry_test::main
             at src/main.rs:11
  15: std::rt::lang_start::{{closure}}
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447/src/libstd/rt.rs:67
  16: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:52
  17: std::panicking::try::do_call
             at src/libstd/panicking.rs:305
  18: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:86
  19: std::panicking::try
             at src/libstd/panicking.rs:281
  20: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  21: std::rt::lang_start_internal
             at src/libstd/rt.rs:51
  22: std::rt::lang_start
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447/src/libstd/rt.rs:67
  23: main
  24: __libc_start_main
  25: _start
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

Process finished with exit code 101
```

domenicquirl:
Early return with an indication that the requested key already existed and a reference back to the value trying to be inserted.

Fixes #90. 

I also took the liberty to add a new `issues.rs` test file to contain reproducible bugs that should now be fixed, so we will know should they appear again.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/91)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -238,7 +238,7 @@
     missing_docs,
     missing_debug_implementations,
     unreachable_pub,
-    intra_doc_link_resolution_failure
+    broken_intra_doc_links
 )]
 #![warn(rust_2018_idioms)]
 #![allow(clippy::cognitive_complexity)]
--- a/src/map.rs
+++ b/src/map.rs
@@ -1815,11 +1815,14 @@ where
                             let current_value = unsafe { current_value.deref() };
 
                             if no_replacement {
-                                // the key is not absent, so don't update
-                                // because of `no_replacement`, we don't use the
-                                // new value, so we need to clean it up
+                                // the key is not absent, so don't update because of
+                                // `no_replacement`, we don't use the new value, so we need to clean
+                                // it up and return it back to the caller
                                 // safety: we own value and did not share it
-                                let _ = unsafe { value.into_owned() };
+                                return PutResult::Exists {
+                                    current: current_value,
+                                    not_inserted: unsafe { value.into_owned().into_box() },
+                                };
                             } else {
                                 // update the value in the existing node
                                 let now_garbage = n.value.swap(value, Ordering::SeqCst, guard);
@@ -1902,11 +1905,14 @@ where
                         // next epoch, which won't arrive until after we drop our guard.
                         let current_value = unsafe { current_value.deref() };
                         if no_replacement {
-                            // the key is not absent, so don't update
-                            // because of `no_replacement`, we don't use the
-                            // new value, so we need to clean it up
+                            // the key is not absent, so don't update because of
+                            // `no_replacement`, we don't use the new value, so we need to clean
+                            // it up and return it back to the caller
                             // safety: we own value and did not share it
-                            let _ = unsafe { value.into_owned() };
+                            return PutResult::Exists {
+                                current: current_value,
+                                not_inserted: unsafe { value.into_owned().into_box() },
+                            };
                         } else {
                             let now_garbage =
                                 tree_node.node.value.swap(value, Ordering::SeqCst, guard);
--- a/tests/regressions.rs
+++ b/tests/regressions.rs
@@ -1,0 +1,18 @@
+use flurry::{epoch::pin, *};
+use rand::{thread_rng, Rng};
+
+#[test]
+fn issue90() {
+    #[cfg(not(miri))]
+    const ITERATIONS: usize = 100_000;
+    #[cfg(miri)]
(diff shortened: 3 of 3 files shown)


domenicquirl:
Interesting. The failure on the warning screening seems to be because `#[deny(intra_doc_link_resolution_failure)]` was renamed?

jonhoo:
Ah, good catch! And :+1: to adding `issues.rs`. Maybe call it `regressions` though?

And yes, the lint has been renamed to `broken_intra_doc_links` (https://github.com/rust-lang/rust/pull/74926).

domenicquirl:
Added the suggested name change and also updated the lint. Interestingly, on my machine I get the warning for the lint the other way round now :D 
Hopefully it's good now on CI.

jonhoo:
Ah, it's probably just because you have a slightly older nightly locally which doesn't have that latest rename PR merged!

domenicquirl:
If it's only changed on nightly than that'll be it, I'm on stable for most things that don't rely on unstable features 👍

domenicquirl:
Everything but Miri ran I think, and Miri timed out last time if I remember correctly - you fine with merging?

jonhoo:
Hmm, miri timing out isn't great. My guess is that it's because this new regression test runs for so many iterations. Iif the test truly _requires_ this many iterations to reproduce the issue, then just use a different number of iterations for miri. You can use `cfg!(miri)` for this. Otherwise, just reduce the number of iterations and miri should be fine.

domenicquirl:
Oh, you think? I thought some other stuff like the stress test were pretty large as well, but seems we already have some miri-dependent tests too. Updating again.

This keeps the original number for non-miri tests, since it runs pretty quickly and the intention is to (hopefully) exercise both the regular and the tree bin path, for which we need some collisions.

jonhoo:
I believe the stress tests are already `#[cfg(not(miri))]`, but could be wrong?

domenicquirl:
They are ignored completely it seems. Other tests have smaller ranges and such. In any case, seems to have been it!

PR #92, opened 2020-08-28
Bump major version of dependencies

jonhoo:
This is a breaking change since `ahash` is public through `DefaultHashBuilder`.

I wonder if we might want to wrap it in a newtype for the future?

This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/92)


ibraheemdev:
Superceded by https://github.com/jonhoo/flurry/pull/105.

jonhoo:
I do still want if we want to hide ahash from our public APIs though 🤔

ibraheemdev:
I don't see why not.

Issue #93, opened 2020-09-21
Miri CI invocation needs updating


RalfJung:
`cargo miri test` changed to be more compatible with `cargo test`, so the old style ([used by this repo](https://github.com/jonhoo/flurry/blob/9a443cbd393c1605eb6df302c0d7531d70eb7be6/azure-pipelines.yml#L60)) of running `cargo miri test -- -Zmiri-ignore-leaks` is deprecated. The new way is `MIRIFLAGS="-Zmiri-ignore-leaks" cargo miri test`.

I'm not familiar with Azure CI or else I'd have prepared a PR. I hope this is not too much of an effort. :)

RalfJung:
I also found [another project of yours which needs a similar update](https://github.com/jonhoo/griddle/blob/4e03134b254e5ed72b3068f9c70e7ce2d55b42c0/azure-pipelines.yml#L48).

jonhoo:
Ah, good catch! Should be easy enough to fix. Just need to replace
```yaml
     - script: cargo miri test -- -Zmiri-ignore-leaks
       displayName: cargo miri test
```
with
```yaml
     - script: cargo miri test
       displayName: cargo miri test
       env:
         MIRIFLAGS: -Zmiri-ignore-leaks
```
I'm pretty swamped these days with thesis writing, but will keep the issue open so I (hopefully) remember when my schedule settles down a little :sweat_smile:

jonhoo:
Closed by #94. Thanks @RalfJung!

PR #94, opened 2020-09-27
fix miri invocation

RalfJung:
Cc https://github.com/jonhoo/flurry/issues/93

I don't know what I am doing here as I never used Azure pipelines.^^

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/94)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,10 +55,10 @@ jobs:
          rust: $(nightly)
          components:
            - miri
-     - script: cargo miri setup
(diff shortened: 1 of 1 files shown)


RalfJung:
It doesn't list the Miri job in PR CI here, no idea why... so I cannot tell if this actually works.

RalfJung:
`cargo miri setup` is also not necessary any more. :)

jonhoo (approved):
Excellent, thank you!

PR #99, opened 2021-11-18
update dependencies and deprecated functions

al8n:
Hi, when I write my project depending on both `flurry` and `crossbeam-epoch`, I meet this error.
```
error[E0308]: mismatched types
   --> src/store/flurry_store.rs:348:26
    |
348 |         map.insert(1, 1, &g);
    |                          ^^ expected struct `flurry::epoch::Guard`, found struct `crossbeam_epoch::Guard`
    |
    = note: expected reference `&flurry::epoch::Guard`
               found reference `&crossbeam_epoch::Guard`
    = note: perhaps two different versions of crate `crossbeam_epoch` are being used?
```

I found some dependencies in this crate that are out-of-date, so I update the dependencies and deprecated functions.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/99)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,17 +20,17 @@ codecov = { repository = "jonhoo/flurry", branch = "master", service = "github" 
 maintenance = { status = "experimental" }
 
 [features]
-sanitize = ['crossbeam-epoch/sanitize']
+#sanitize = ['crossbeam-epoch/sanitize']
 
 [dependencies]
-crossbeam-epoch = "0.8.2"
-parking_lot = "0.10"
-num_cpus = "1.12.0"
-rayon = {version = "1.3", optional = true}
+crossbeam-epoch = "0.9"
+parking_lot = "0.11"
+num_cpus = "1.13"
+rayon = {version = "1.5", optional = true}
 serde = {version = "1.0.105", optional = true}
 
 [dependencies.ahash]
-version = "0.3.2"
+version = "0.7.6"
 default-features = false
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "flurry"
 version = "0.3.1"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
-edition = "2021"
+edition = "2018"
 license = "MIT OR Apache-2.0"
 
 readme = "README.md"
@@ -18,9 +18,6 @@ exclude = ["/jsr166/**"]
 azure-devops = { project = "jonhoo/jonhoo", pipeline = "flurry", build = "15" }
 codecov = { repository = "jonhoo/flurry", branch = "master", service = "github" }
 maintenance = { status = "experimental" }
-
-[features]
-#sanitize = ['crossbeam-epoch/sanitize']
 
 [dependencies]
 crossbeam-epoch = "0.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "flurry"
 version = "0.3.1"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
-edition = "2018"
+edition = "2021"
 license = "MIT OR Apache-2.0"
 
 readme = "README.md"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
   - template: default.yml@templates
     parameters:
-      minrust: 1.56.0 # std::hint::spin_loop
+      minrust: 1.49.0 # std::hint::spin_loop
   - job: deny
     displayName: "Disallowed attributes"
     pool:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,103 +1,103 @@
 jobs:
- - template: default.yml@templates
-   parameters:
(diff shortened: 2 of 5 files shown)


al8n:
Hi, @jonhoo. I have no idea why the tarpaulin test failed, could you give a review? Thanks!

azure-pipelines.yml
```diff
@@ -1,103 +1,103 @@
...
-       displayName: Enable debug symbols
-     - script: |
-           env RUSTFLAGS="-Z sanitizer=leak" cargo test --all-features --target x86_64-unknown-linux-gnu
-       displayName: cargo -Z sanitizer=leak test
- - template: coverage.yml@templates
-   parameters:
-     token: $(CODECOV_TOKEN_SECRET)
-     nightly: true
-     doctests: true
+  - template: default.yml@templates
```
jonhoo, line 2: Why all of these whitespace changes?

Cargo.toml
```diff
@@ -20,17 +20,17 @@ codecov = { repository = "jonhoo/flurry", branch = "master", service = "github"
 maintenance = { status = "experimental" }
 
 [features]
-sanitize = ['crossbeam-epoch/sanitize']
+#sanitize = ['crossbeam-epoch/sanitize']
```
jonhoo, line 23: Did this feature actually get removed, or just renamed? I forget whether we make use of `sanitize` directly in the code of this crate, but if we do this should now be
```toml
sanitize = []
```
If we don't, the feature should be removed entirely, and you'll also need to update the CI config to not try to run with `--features sanitize`.

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "flurry"
 version = "0.3.1"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
-edition = "2018"
+edition = "2021"
```
jonhoo, line 5: This would bump the minimum supported rust version of the crate, which I don't think is warranted quite yet. I'd like to leave this as-is for now.

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "flurry"
 version = "0.3.1"
```
jonhoo, line 3: This is definitely a breaking change, but I'm okay with that. Just leaving this as a reminder to myself.

src/raw/mod.rs
```diff
@@ -81,10 +81,11 @@ impl<K, V> Table<K, V> {
             t if t.is_null() => {
                 // if a no next table is yet associated with this table,
                 // create one and store it in `self.next_table`
-                match self.next_table.compare_and_set(
+                match self.next_table.compare_exchange(
                     Shared::null(),
                     for_table,
                     Ordering::SeqCst,
+                    Ordering::SeqCst,
```
jonhoo, line 88:
```suggestion
                    Ordering::Relaxed,
```

src/raw/mod.rs
```diff
@@ -311,12 +312,12 @@ impl<K, V> Table<K, V> {
...
     ) -> Result<
         Shared<'g, BinEntry<K, V>>,
-        crossbeam_epoch::CompareAndSetError<'g, BinEntry<K, V>, P>,
+        crossbeam_epoch::CompareExchangeError<'g, BinEntry<K, V>, P>,
     >
     where
         P: Pointer<BinEntry<K, V>>,
     {
-        self.bins[i].compare_and_set(current, new, Ordering::AcqRel, guard)
+        self.bins[i].compare_exchange(current, new, Ordering::AcqRel, Ordering::Acquire, guard)
```
jonhoo, line 320:
```suggestion
        self.bins[i].compare_exchange(current, new, Ordering::AcqRel, Ordering::Relaxed, guard)
```

tests/cuckoo/stress.rs
```diff
@@ -102,7 +105,10 @@ fn stress_delete_thread(env: Arc<Environment>) {
     while !env.finished.load(Ordering::SeqCst) {
         let idx = env.ind_dist.sample(&mut rng);
         let in_use = env.in_use.lock();
-        if (*in_use)[idx].compare_and_swap(false, true, Ordering::SeqCst) {
+        if (*in_use)[idx]
+            .compare_exchange_weak(false, true, Ordering::SeqCst, Ordering::SeqCst)
```
jonhoo, line 109: Why did you make this use `_weak`?
```suggestion
            .compare_exchange(false, true, Ordering::SeqCst, Ordering::Relaxed)
```

tests/cuckoo/stress.rs
```diff
@@ -125,7 +131,10 @@ fn stress_find_thread(env: Arc<Environment>) {
...
```
jonhoo, line 135:
```suggestion
            .compare_exchange(false, true, Ordering::SeqCst, Ordering::Relaxed)
```

jonhoo (changes requested):
Thanks! A couple of mostly smaller notes.

azure-pipelines.yml
al8n: That is my bad, my editor auto-reformat the yml file.

Cargo.toml
al8n: Yes, the feature is removed by crossbeam-epoch.

al8n:
Hi, jonhoo. Thanks for your suggestions, all suggestions have been applied now.

Issue #11, opened 2020-01-23
Implement the sharded counters optimization


jonhoo:
[The Java code](https://github.com/jonhoo/flurry/blob/d3dae0465b37b7f12c4f0d58a16f36fb1d8c1596/jsr166/src/ConcurrentHashMap.java#L400-L411) uses the `LongAdder` and`CounterCell` types in `add_count` to both improve the scalability of tracking the number of elements, and to intelligently sense when a resize should happen. We should port that code over as well.

PR #100, opened 2021-11-23
Shared counters optimization #11

jimvdl:
Hello!

I've made an attempt at implementing the counter cell optimization. All of the tests currently pass, however this doesn't mean what I wrote works. I'm unsure how to test all branches when `count` experiences contention. 

I looked at the Java implementation and they use `ThreadLocalRandom`, as a counterpart I opted for `SmallRng`. According to the Java documentation it's not cryptographically secure, neither is `SmallRng`. Most of [`rand`'s generators](https://docs.rs/rand/0.8.4/rand/rngs/index.html#our-generators) are secure except for `StdRng` and `SmallRng`.

Also, in the Java version they initialize the array of `CounterCell`'s when `baseCount` experiences contention, however I figured the cost of initializing the array wouldn't be that high so it defaults to `vec![AtomicIsize::new(0), AtomicIsize::new(0)]` in `HashMap::with_hasher`.  

I'm opening this as a draft because ~~I might need to add tests (although I'm not sure how to test this)~~ a lot of the tests detect memory leaks (the capacity of the map isn't correct.) and I probably made plenty of mistakes. (this is my first open-source contribution)

Any feedback is greatly appreciated.

Fixes: #11 

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/100)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ sanitize = ['crossbeam-epoch/sanitize']
 crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
+rand = { version = "0.7", features = ["small_rng"] }
 rayon = {version = "1.3", optional = true}
 serde = {version = "1.0.105", optional = true}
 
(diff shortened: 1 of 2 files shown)


jimvdl:
I've just figured out that I was not running the tests with `cargo test --all-features`, sorry about that.
I'll close this draft for now until I actually get all the tests to pass.

Issue #98, opened 2021-09-27
Unsoundness in `HashMap::clear`


ibraheemdev:
`HashMap::clear` does not have `K\V: Send + Sync` bounds, even though it calls `defer_destroy` on nodes and values, which might call the destructors on another thread. Unless I am missing something, this is unsound and could cause undefined behavior.

jonhoo:
Ah, yes, you are entirely right! Good catch. Would you mind submitting a PR with a fix?

PR #101, opened 2021-11-27
Shared counters optimization

jimvdl:
Hello, I've made an attempt at implementing the shared counters optimization. 

Overview:
Once `count` experiences contention the compare exchange will fail triggering the allocation of `counter_cells`. It allocates a `Vec` with 2 cells initializing one of them to `n`. Any future contention on `count` will randomly select a counter cell and increment its count. If this also fails (and the number of counter cells doesn't exceed the number of cpu's) then a new counter cell array is allocated with n << 1, it copies all of the old values over and initializes the new counters to 0. If `counter_cells` is busy for whatever reason it falls back on simply attempting to increment count again. When the hash map is dropped the `counter_cells` array gets subsequently dropped.

Fixes: #11 

This is my first open-source contribution so any advice/feedback is greatly appreciated.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/101)


jimvdl:
Hi @jonhoo, would you mind giving this a review? Thanks!

jonhoo:
This is on my radar, I just haven't had the spare time to look at it yet! Thanks for giving it a shot. Once I have some time I'll give it a review :)

One thing off the top of my head is that I'd like to see this implemented as a separate type in its own module rather than all inlined into the implementation. Will make it much easier to review, and possibly reusable in and of itself!

jimvdl:
I had some questions, I'm sure you'll get to them naturally when you're reviewing but wanted to list them out:
- The struct name is currently `LongAdder`, I'm not sure if that is a fitting name since we are not adding longs like in Java. (maybe IsizeAdder, but I'll leave that up to you).
- A bunch of the tests already verify that the length matches up but since the counter is not a atomic snapshot some tests might occasionally fail due to the sum being calculated while asserting the length.
- Should I directly mirror LongAdder and implement functions like increment(), decrement() etc?
- The flurry style linting beta clippy test warns about unsoundness on `K` & `V` on `BinEntry` because they do not implement `Send` and `Sync`, should I PR a fix for these before this PR gets reviewed? 

Take all the time you need, I'm eager to learn something and potentially get this merged so we'll take a look at it whenever you're ready!

jonhoo:
> I had some questions, I'm sure you'll get to them naturally when you're reviewing but wanted to list them out:

I didn't get to as many things as I wanted today, so this remains on my backlog, but figured I'd at least answer your questions!

>     * The struct name is currently `LongAdder`, I'm not sure if that is a fitting name since we are not adding longs like in Java. (maybe IsizeAdder, but I'll leave that up to you).

Oof, `Isize` just looks _really_ weird. How about `ConcurrentCounter`? That's _exactly_ what it is. I suppose we could throw "signed" in there, but I don't think that's super necessary.

>     * A bunch of the tests already verify that the length matches up but since the counter is not a atomic snapshot some tests might occasionally fail due to the sum being calculated while asserting the length.

We _definitely_ do not want spurious tests, but it sounds like those tests should be racy already if they're seeing a race when using this new concurrent counter... Can you point to a specific test that fails?

>     * Should I directly mirror LongAdder and implement functions like increment(), decrement() etc?

I wonder — could you just implement the `AddAssign` trait instead? I think your `add` only takes a `Guard` at the moment because it uses `Atomic`, but maybe it'd be an idea to just allocate a `Vec` with `num_cpu` counters from the very beginning and keep a _separate_ `AtomicUsize` that just tracks the highest index in the `Vec` we should be using? That means no resizing, which means no `Atomic`, which means no `Guard`. Want to give that a shot?

>     * The flurry style linting beta clippy test warns about unsoundness on `K` & `V` on `BinEntry` because they do not implement `Send` and `Sync`, should I PR a fix for these before this PR gets reviewed?

Oh, interesting. Yeah, definitely try for a separate PR that adds those!

> Take all the time you need, I'm eager to learn something and potentially get this merged so we'll take a look at it whenever you're ready!

:heart:

jimvdl:
> ```
> We _definitely_ do not want spurious tests, but it sounds like those tests should be racy already if they're seeing a race when using this new concurrent counter... Can you point to a specific test that fails?
> ```
Every test that asserts the maps length might be racy. Before the counter optimization the length of the map would always be correctly synced. Now however, when high contention occurs the length might be a bit behind the actual number of items in the map. When a test asserts the length while it's still counting the length might be 1 behind the actual length (if that makes sense). I did try to reproduce tests that might be spurious but couldn't find any as of yet. 

>```
> I wonder — could you just implement the `AddAssign` trait instead? I think your `add` only takes a `Guard` at the moment because it uses `Atomic`, but maybe it'd be an idea to just allocate a `Vec` with `num_cpu` counters from the very beginning and keep a _separate_ `AtomicUsize` that just tracks the highest index in the `Vec` we should be using? That means no resizing, which means no `Atomic`, which means no `Guard`. Want to give that a shot?
> ```
I tried to implement the `AddAssign` trait but got slightly stuck due to the trait requiring `&mut self`. `add_count` only has `&self` and something like `RefCell` probably wouldn't work here due to it not being `Sync`. If we somehow can get around the exclusive borrow requirement `AddAssign` should work. Any tips/thoughts on how I might go about this?

> ```
> Oh, interesting. Yeah, definitely try for a separate PR that adds those!
> ```
Will do. I'll also include #98 since it's a similar problem.

src/counter.rs
```diff
@@ -0,0 +1,63 @@
...
+            .base
+            .compare_exchange(base, base + value, Ordering::SeqCst, Ordering::Relaxed)
+            .is_err()
+        {
+            // we experienced contention on base
+            loop {
+                let c = &self
+                    .cells
+                    .choose(&mut rand::rngs::SmallRng::from_entropy())
+                    .unwrap();
```
jonhoo, line 33: This is maybe a super weird proposal, but what if we chose a cell based on `base % self.cells.len()` (maybe expressed better since we know `self.cells` is base-2-length? That way we don't need to sample randomness at all.

src/counter.rs
```diff
@@ -0,0 +1,63 @@
...
+                let cv = c.load(Ordering::SeqCst);
+                if c.compare_exchange(cv, cv + value, Ordering::SeqCst, Ordering::Relaxed)
+                    .is_ok()
+                {
+                    // we've successfully incremented a random counter
+                    break;
+                }
+
+                // the selected counter also experienced contention, retry base
```
jonhoo, line 42: This makes it seem like we should just have the `loop` span trying both `base` _and_ a cell? That way we won't need to repeat the logic for base here.

src/counter.rs
```diff
@@ -0,0 +1,63 @@
...
+
+                // both base and a random cell experienced contention, retry another cell
+            }
+        }
+    }
+
+    pub(crate) fn sum(&self) -> isize {
+        let sum: isize = self.cells.iter().map(|c| c.load(Ordering::SeqCst)).sum();
+
+        self.base.load(Ordering::SeqCst) + sum
```
jonhoo, line 61: I actually wonder if these should be `Relaxed` loads instead, since they're just using for counting. Or maybe `sum` should take the `Ordering` to use as an argument so the caller can decide?

src/map.rs
```diff
@@ -1161,21 +1168,21 @@ where
     }
 
     fn add_count(&self, n: isize, resize_hint: Option<usize>, guard: &Guard) {
-        // TODO: implement the Java CounterCell business here
```
jonhoo, line 1164: There's a _lot_ more to [the real Java implementation of `CounterCell`](https://github.com/jonhoo/flurry/blob/13db2fb33d0c7ab45205ac5bb00d9b8749d959a2/jsr166/src/ConcurrentHashMap.java#L2295). We don't need to do all of that here in the current PR of course, but may be worth keeping a TODO around to point at the work that still remains.

src/map.rs
```diff
@@ -1161,21 +1168,21 @@ where
...
-            cmp::Ordering::Less => self.count.fetch_sub(n.abs(), Ordering::SeqCst) - n,
-            cmp::Ordering::Equal => self.count.load(Ordering::SeqCst),
-        };
+        self.counter.add(n);
 
         // if resize_hint is None, it means the caller does not want us to consider a resize.
         // if it is Some(n), the caller saw n entries in a bin
-        if resize_hint.is_none() {
-            return;
+        match resize_hint {
```
jonhoo, line 1175: I'm not sure I follow why this had to change? I thought we did as the same as the Java implementation here, which also has the same counter semantics as what we're now implementing.

jonhoo (changes requested):
I was finally able to give this a look, sorry it took so so so long!

@ibraheemdev Any chance you have some spare cycles to try benchmarking this to see if it meaningfully improves multi-core performance? Or better yet, help @jimvdl do the benchmarks themselves!

jonhoo:
Oh, and you'll probably want to merge your changes with master, since some things have changed there in the intervening time.

As for `AddAssign`, I think you could probably implement it for `&Counter`, but it's not terribly important, and may honestly just end up proving unexpected and unergonomic in the end :sweat_smile:

ibraheemdev:
> Any chance you have some spare cycles to try benchmarking this to see if it meaningfully improves multi-core performance?

I may have some time next weekend. It also occurs to me that this datastructure may be generally useful as a crate.

src/map.rs
jimvdl: When looking at [this line](https://github.com/jonhoo/flurry/blob/13db2fb33d0c7ab45205ac5bb00d9b8749d959a2/jsr166/src/ConcurrentHashMap.java#L2308) I interpreted `Some(0)` and `Some(1)` as also signifying that no resize should be considered?

src/counter.rs
jimvdl: Good idea, taking the `Ordering` is a nice way to get best of both worlds.

src/counter.rs
jimvdl: I think it's actually a nicer way of selecting a cell without `rand`, with the added bonus that we can remove that dependency.

jimvdl:
> As for `AddAssign`, I think you could probably implement it for `&Counter`

I think your hunch might be correct, it would mean you could increment the counter like this:
```rust
let mut c = &self.counter;
c += 5;
```
Which does feel a bit unergonomic, but I'm still struggling with some of Rust's basics so maybe there is a nicer way.

src/counter.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    pub(crate) fn add(&self, value: isize) {
+        let base = self.base.load(Ordering::SeqCst);
+
+        while self
+            .base
+            .compare_exchange(base, base + value, Ordering::SeqCst, Ordering::Relaxed)
+            .is_err()
+        {
+            let c = &self.cells[base as usize % self.cells.len()];
```
jonhoo, line 27: Actually, thinking some more on this, I wonder if this will just make all the threads that fail to increment `base` then decide to use the _same_ cell. We could use `base + value`, but even that's not great since `value` is _usually_ -1 or 1. I wonder if there's something else we can seed it based off of. I also wonder if we should change the index on each iteration (currently it ends up being loop-invariant). Maybe by mixing in the value we read from `cv`?

src/map.rs
```diff
@@ -103,6 +101,8 @@ pub struct HashMap<K, V, S = crate::DefaultHashBuilder> {
     /// next element count value upon which to resize the table.
     size_ctl: AtomicIsize,
 
+    counter: ConcurrentCounter,
```
jonhoo, line 104: Any particular reason why you moved where this field is in the struct?

src/map.rs
```diff
@@ -906,7 +912,7 @@ where
 
                     next_table.store_bin(i, low_bin);
                     next_table.store_bin(i + n, high_bin);
-                    table.store_bin(i, table.get_moved(Shared::from(next_table_ptr), guard));
+                    table.store_bin(i, table.get_moved(next_table_ptr, guard));
```
jonhoo, line 915: Hmm, this change doesn't seem right. Did the merge go wrong somewhere maybe?

src/map.rs
```diff
@@ -1141,21 +1147,22 @@ where
...
     fn add_count(&self, n: isize, resize_hint: Option<usize>, guard: &Guard<'_>) {
-        // TODO: implement the Java CounterCell business here
-
-        use std::cmp;
-        let mut count = match n.cmp(&0) {
-            cmp::Ordering::Greater => self.count.fetch_add(n, Ordering::SeqCst) + n,
-            cmp::Ordering::Less => self.count.fetch_sub(n.abs(), Ordering::SeqCst) - n,
-            cmp::Ordering::Equal => self.count.load(Ordering::SeqCst),
-        };
+        // TODO: finish Java CounterCell port, this is only a bare minimum implementation.
```
jonhoo, line 1150: I wonder if this comment should go _inside_ `CounterCell` rather than be just here where we use it?

src/map.rs
jonhoo: Oh, huh, yeah, you're right. How about then `resize_hint.unwrap_or(0) <= 1`?

jonhoo (changes requested):
Yeah, it's not super pretty. I think it's okay just leaving that out then — it's not like calling `.add` is _that_ onerous!

src/map.rs
jimvdl: Oh I normally use the formatter in vscode and it probably moved it for whatever reason, sorry about that.

src/map.rs
jimvdl: When I ran `cargo clippy` it came up with the following:
```
warning: useless conversion to the same type: `reclaim::Shared<raw::Table<K, V>>`
   --> src\map.rs:915:56
    |
915 |                     table.store_bin(i, table.get_moved(Shared::from(next_table_ptr), guard));
    |                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `Shared::from()`: `next_table_ptr`
    |
    = note: `#[warn(clippy::useless_conversion)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
```
I can revert the changes because I do realise that removing those conversions in this PR might be out of place.

jimvdl:
I'm also not sure how to accurately test the map's length, because it might not always be in sync with the actual number of items. See the [Java docs](https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/java/util/concurrent/atomic/LongAdder.java#L110). 

I glanced at their tests to see if they have a special way of asserting the length but couldn't find anything, maybe I missed something? What do you guys think?

src/map.rs
jonhoo: Oh, yeah, let me fix that real quick just directly on `master` :+1:

src/counter.rs
jonhoo: I think that'd be a great thing to try and port! But for the purposes of this PR, that _may_ be overkill unless you specifically want to do it :) In the absence of that though, we do need to use a better index than _just_ `base` since all the contending threads will pick the same cell. I think `base + value` that also gets `+= cv` on each loop is a good place to start.

jonhoo:
> I'm also not sure how to accurately test the map's length

Hmm, how about something like
```rust
loop {
  let n = map.len()
  assert!(n <= EXPECTED, "n <= {}", EXPECTED);
  if n == EXPECTED { break; }
  std::thread::yield_now();
}
```

Maybe even with a max limit for the number of iterations? Probably worth sticking that in some kind of helper function for the tests too.

src/map.rs
jonhoo: Fixed on `master`.

src/map.rs
jimvdl: Looking at `if check <= 1` should only be executed if there are counter cells allocated and it couldn't update base due to contention. The previous `if resize_hint.is_none()` worked better because now I'm seeing some weird behaviour where the table is not resizing when it should. For example the `map::capacity()` test fails due to this change.

jimvdl:
Sorry that I keep bothering you with this but I'm seriously confused. I've been re-running the tests over and over to try and catch a test that fails due to the length being off by 1 (like I saw way before) but can't, for the life of me, find one that fails now... Before it would fail pretty often but ever since we changed the cells the problem _seems_ to have gone away?? 

Since I can't reproduce the problem anymore _at all_ we probably won't need that helper function.

Also, when I was testing I tried a sample size of 1 million parallel inserts and got this distribution: `ConcurrentCounter { base: 990579, cells: [2384, 2308, 2360, 2369] }`. On the one hand the solution of `base + value` and each iteration of the loop adding `+= cv` works pretty well, on the other hand it seems that you need to do a ridiculous amount of parallel inserts to make use of the counter cells. We can still add some benchmarks to test this a little better, although it seems consistent.

jonhoo:
Huh, how weird. Maybe the old implementation had a bug somehow? It's a good sign it's not happening frequently now though!

That distribution looks pretty good. I think benchmarks will _definitely_ be helpful here though, and hopefully we'll see an impact at higher thread counts.

jonhoo (approved):

jimvdl:
I'll soon give adding benchmarks for this a try when I have a little bit more time. I did open a PR for the benchmarks as they currently do not compile, once that's resolved I'll get back to this PR.

jonhoo:
Hmm, that's interesting. If I'm reading your data correctly, it seems like `ConcurrentCounter` is _slower_ under contention, not faster as we'd expect :thinking: How many (real) cores do you have on the computer you ran this on?

jimvdl:
I ran this on an Intel Core i7-7700K, which has 4 cores and 8 threads. These benchmarks don't even take the actual insert operation into account, which makes me suspect that the impact of having just one `AtomicIsize` would even be less noticable in practise.

jonhoo:
Yeah, it's tricky, because where I think this would matter is when you have, say, 16, or 32 real cores, which is harder to test for. I wonder what results you'd get on your box if you ran with 4 threads though — hyperthreads tend to only really add noise for these kinds of measurements.

jimvdl:
I will run the benchmarks again when I use only 4 cores to see if that makes a difference, I'll get back to you on that one. 

I _might_ be able to test this on a CPU with 10 cores, and see if that makes a difference. Although let's say for a moment that having 16 or 32 makes a substantial difference, would the `ConcurrentCounter` still be useful knowing that with fewer than 16 cores you would get a performance decrease?

jimvdl:
I've tried to disable hyperthreading but apparently my CPU doesn't have an option to disable it. (as far as I can find) I did try another route: in the task manager you can set the affinity of a running process, which I then set to 4 instead of the 8 it had before. I'm not sure if this gave me accurate results but here they are:

`AtomicIsize` at 1 thread:
  | Lower bound | Estimate | Upper bound
-- | -- | -- | --
Slope | 129.12 us | 129.23 us | 129.36 us
Throughput | 253.31 Melem/s | 253.56 Melem/s | 253.78 Melem/s
R² | 0.9981254 | 0.9982359 | 0.9981086
Mean | 129.14 us | 129.26 us | 129.40 us
Std. Dev. | 379.87 ns | 682.82 ns | 953.36 ns
Median | 128.98 us | 129.03 us | 129.13 us
MAD | 207.45 ns | 279.07 ns | 362.17 ns

`ConcurrentCounter` at 1 thread:
  | Lower bound | Estimate | Upper bound
-- | -- | -- | --
Slope | 231.46 us | 231.60 us | 231.77 us
Throughput | 141.38 Melem/s | 141.49 Melem/s | 141.57 Melem/s
R² | 0.9991705 | 0.9992210 | 0.9991435
Mean | 231.51 us | 231.66 us | 231.85 us
Std. Dev. | 409.77 ns | 891.91 ns | 1.2775 us
Median | 231.42 us | 231.47 us | 231.53 us
MAD | 212.07 ns | 293.64 ns | 343.09 ns

`AtomicIsize` at 4 threads:
  | Lower bound | Estimate | Upper bound
-- | -- | -- | --
Slope | 455.77 us | 477.68 us | 498.68 us
Throughput | 65.709 Melem/s | 68.598 Melem/s | 71.896 Melem/s
R² | 0.2564080 | 0.2696591 | 0.2574293
Mean | 449.16 us | 464.91 us | 480.76 us
Std. Dev. | 76.443 us | 80.814 us | 84.013 us
Median | 407.94 us | 413.40 us | 528.11 us
MAD | 55.867 us | 67.139 us | 128.89 us

`ConcurrentCounter` at 4 theads:
  | Lower bound | Estimate | Upper bound
-- | -- | -- | --
Slope | 742.61 us | 756.28 us | 769.56 us
Throughput | 42.580 Melem/s | 43.328 Melem/s | 44.125 Melem/s
R² | 0.5678183 | 0.5859128 | 0.5687718
Mean | 745.44 us | 753.67 us | 761.84 us
Std. Dev. | 35.724 us | 42.078 us | 48.248 us
Median | 746.76 us | 751.40 us | 759.00 us
MAD | 24.176 us | 41.755 us | 54.862 us


If you compare these to the 4 thread versions of the previous benchmark with hyperthreading enabled you can see a slight difference:

`AtomicIsize` with 4 threads: 
  | Lower bound | Estimate | Upper bound
-- | -- | -- | --
Slope | 485.85 us | 496.67 us | 506.66 us
Throughput | 64.674 Melem/s | 65.975 Melem/s | 67.444 Melem/s
R² | 0.5423258 | 0.5583073 | 0.5446181
Mean | 481.40 us | 490.08 us | 498.65 us
Std. Dev. | 40.887 us | 44.328 us | 46.723 us
Median | 470.72 us | 520.55 us | 523.34 us
MAD | 11.299 us | 16.970 us | 67.557 us

`ConcurrentCounter` with 4 threads:
  | Lower bound | Estimate | Upper bound
-- | -- | -- | --
Slope | 1.0255 ms | 1.0359 ms | 1.0465 ms
Throughput | 31.313 Melem/s | 31.632 Melem/s | 31.954 Melem/s
R² | 0.8311427 | 0.8397268 | 0.8309221
Mean | 1.0211 ms | 1.0318 ms | 1.0422 ms
Std. Dev. | 44.870 us | 54.107 us | 63.172 us
Median | 1.0214 ms | 1.0305 ms | 1.0466 ms
MAD | 37.691 us | 49.023 us | 59.846 us

jonhoo:
Hmm, that still looks like it's slower _and_ less scalable, which is surprising. I'd be super curious to see on a higher-core-count machine, and ideally as a plot with error bars!

jimvdl:
I tried to get my hands on that 10 core machine but sadly couldn't use it. Anything else I can try/do?

jonhoo:
@ibraheemdev Any chance you have a box with more cores available? I might be able to spin something up, but my schedule is pretty packed for a while :disappointed:

ibraheemdev:
My machine has 8 cores, I can run the benchmarks but not sure we'll see any benefit at a low-medium core count.

JackThomson2:
Hey was interested by this PR and had a play around, I found using the nightly only `#[thread_local]` gave some promising results
```
jack_counter/1          time:   [158.47 us 159.09 us 160.13 us]
                        thrpt:  [204.63 Melem/s 205.97 Melem/s 206.77 Melem/s]
```

```
jack_counter/4          time:   [519.68 us 526.90 us 533.70 us]
                        thrpt:  [61.398 Melem/s 62.190 Melem/s 63.054 Melem/s]
```

```
jack_counter/8          time:   [421.63 us 427.01 us 432.60 us]
                        thrpt:  [75.746 Melem/s 76.738 Melem/s 77.718 Melem/s]
```

Compared to the AtomicIsize 

```
atomic_counter/1        time:   [156.52 us 156.65 us 156.83 us]
                        thrpt:  [208.95 Melem/s 209.18 Melem/s 209.35 Melem/s]
```

```
atomic_counter/4        time:   [575.31 us 576.61 us 577.85 us]
                        thrpt:  [56.707 Melem/s 56.828 Melem/s 56.958 Melem/s]
```

```
atomic_counter/8        time:   [592.14 us 592.63 us 593.18 us]
                        thrpt:  [55.241 Melem/s 55.293 Melem/s 55.338 Melem/s]
```

This was using the benchmarking setup from this PR and I have the code here if you want to have a look: https://github.com/JackThomson2/fast-counter/blob/master/src/lib.rs

jimvdl:
Those results are promising indeed! I like the solution you went with, definitely worth looking into more. I'm wondering why the thread local approach is more performant compared to my solution. If you possible published this as a crate it could be used in Flurry as a dependency instead.

JackThomson2:
I'll have a look at getting this published / compare the performance to the thread_local non nightly macro. 

I originally had a look at optimising your approach, the few findings I could see which helped speed it up where
 - Ensuring the number of cpus where set to next_power_of_two() and eliminating the divide here `let c = &self.cells[index as usize % self.cells.len()];`
 - Adding another  `base = self.base.load(Ordering::SeqCst);` at the end of the loop (line 48) this meant it was less likely to fail, I assumed under higher contention this number will have been updated by the time we came around.

I think the reason it's not as fast was that you just moved the contention to the next cell. I had an experiment using wyrand as a pseudo random generator to pick a random cell, and this did also help slightly

JackThomson2:
I was benchmarking this on my 16 core cpu, seems to scale really well as the thread count goes up I'll upload the results soon.

jimvdl:
All the results look way better than a single atomic counter. It has some additional overhead for 2 core cpu's but honestly I don't think that would be an issue since everyone has at least 2 cores anyway.

You want to open a PR for this instead? If you do lmk I'll close this one.

JackThomson2:
Even better news my suspicion around the inlining was correct was correct, when I added `#[inline]` we're much closer on 2 core and faster on 4 core!

```
atomic_counter/2        time:   [290.27 us 293.65 us 297.26 us]
                        thrpt:  [110.23 Melem/s 111.59 Melem/s 112.89 Melem/s]

atomic_counter/4        time:   [320.62 us 323.01 us 325.27 us]
                        thrpt:  [100.74 Melem/s 101.45 Melem/s 102.20 Melem/s]

atomic_counter/8        time:   [343.33 us 344.14 us 344.98 us]
                        thrpt:  [94.985 Melem/s 95.217 Melem/s 95.442 Melem/s]

atomic_counter/16       time:   [410.49 us 411.71 us 412.99 us]
                        thrpt:  [79.344 Melem/s 79.590 Melem/s 79.827 Melem/s]

------------------------------------------------------------------------------

fast_counter_nightly/2  time:   [314.05 us 315.63 us 317.16 us]
                        thrpt:  [103.32 Melem/s 103.82 Melem/s 104.34 Melem/s]

fast_counter_nightly/4  time:   [292.82 us 294.93 us 296.72 us]
                        thrpt:  [110.44 Melem/s 111.10 Melem/s 111.91 Melem/s]

fast_counter_nightly/8  time:   [209.61 us 215.30 us 221.28 us]
                        thrpt:  [148.08 Melem/s 152.20 Melem/s 156.33 Melem/s]

fast_counter_nightly/16 time:   [157.28 us 160.06 us 163.00 us]
                        thrpt:  [201.04 Melem/s 204.72 Melem/s 208.34 Melem/s]

------------------------------------------------------------------------------


fast_counter_stable/2   time:   [400.89 us 407.77 us 413.33 us]
                        thrpt:  [79.277 Melem/s 80.360 Melem/s 81.739 Melem/s]

fast_counter_stable/4   time:   [369.10 us 372.90 us 376.90 us]
                        thrpt:  [86.942 Melem/s 87.873 Melem/s 88.778 Melem/s]

fast_counter_stable/8   time:   [247.36 us 253.10 us 258.51 us]
                        thrpt:  [126.76 Melem/s 129.47 Melem/s 132.47 Melem/s]

fast_counter_stable/16  time:   [162.17 us 166.01 us 170.13 us]
                        thrpt:  [192.60 Melem/s 197.39 Melem/s 202.06 Melem/s]
```

I'll see if I can get time to make the PR to change to this!

jimvdl:
Closing in favour of #109.

jonhoo:
Great work folks, thanks for picking this up and driving it on your own!

PR #95, opened 2020-10-22
Move epoch garbage collection to flize

domenicquirl:
After chatting with Acrimon on Discord, we added the missing features (mostly unprotected shields) to `flize`, making it possible to replace `crossbeam_epoch` with it in `flurry`. The main motivation behind this change is to overcome some of our performance issues that seem to be related to garbage collection (see #50, https://github.com/jonhoo/bustle/issues/2). There's some amount of stuff going on here, but I tried to keep the code as familiar as possible. Preliminary evaluation:

## Disadvantages
  - **Type/Trait Complexity & Lifetimes:** there are (obviously) some differences in the interface the two libraries  provide. `flize`'s `Shield` is a trait with an associated lifetime bound, which permeates into a lot of our types and functions now, making it harder to understand and maintain them (in particular for new contributors). For example, `NodeIter<'g, K, V>` is now `NodeIter<'m, 'g, K, V, SH>`, where `SH: flize::Shield<'m>`. Or, for a function example, my now favourite function (guard lifetimes elided):
  ```rust
    pub(crate) fn guarded_eq<'m1, 'm2, SH, SH2>(
        &'m1 self,
        other: &'m2 Self,
        our_guard: &Guard<'m1, SH>,
        their_guard: &Guard<'m2, SH2>,
    ) -> bool
    where
        V: PartialEq,
        SH: flize::Shield<'m1>,
        SH2: flize::Shield<'m2>
  ```
  - **Shared Collectors:** it is no longer possible to share a collector with other data structures, since `flize` doesn't have the concept of a global default collector.
    * Because of this, `MapRef` and `SetRef` cannot be constructed `with_guard` anymore 
  - We now have our own small `ebr` module to extend `flize` to our needs, which is some additional code to maintain
  - I've had to write a lot of manual `Debug` impls (though that can be fixed very easily by implementing `Debug` for `flize` types, mainly atomics)

## Advantages
  - **No `Owned` Type:** `flize` only uses `Atomic` and `Shared`. This removes _a lot_ of `Owned::new(...).into_shared(&guard)`, which almost always happens together in `flurry`, and has proven to be quite ergonomic.
  - **Collector safety:** The PR keeps the collecor/guard check on map access to ensure guards from the correct collector are used (`disallow_evil`).
  - **Performance:** `flize` is _fast_, which means `flurry` is now also fast! The map now has usable performance comparable to other concurrent maps (see bench results below)

[...]

PR #102, opened 2022-02-09
Move garbage collection to seize

ibraheemdev:
## Background

This PR moves garbage collection from `crossbeam-epoch` to `seize`. As shown in #50, crossbeam-epoch has serious performance issues that make flurry much slower than other concurrent hashmaps. #95 attempted to switch to `flize`, but unfortunately the pull request as well as the library have stalled.

## Design

`seize` uses a special type of reference counting instead of tracking global epochs. When an object is retired, it is first appended to a thread-local batch. This operation does not allocate. When the batch size hits a configurable limit, the batch is added to a separate list for each thread that is currently active (holds a guard). The reference count of the batch is set to the number of threads that are active. After every thread drops its guard, it decrements the reference count. If it holds the last reference, it frees the entire batch.

## Implementation

The API provides by seize differs quite a bit from that of crossbeam-epoch. I tried to make migration easier by adding a `reclaim` module that mimics some of the previously used APIs, but here are some of the major changes:

- `seize` requires metadata to be stored with every allocated object, so `T -> Linked<T>` has propagated around the codebase.
- `seize` has no concept of a global collector, which means that:
  - The free-function `pin` has been removed.
  - `with_collector` is now public as the footgun of using `epoch::pin()` no longer exists.
  - `Guard` has a lifetime. Public functions now take `&'g Guard<'_>` instead of `&'g Guard`.
  - Many internal functions now need access to the `&Collector` in order to allocate objects.

## Performance

This algorithm has shown to have performance similar to or better than optimized epoch-based schemes (like flize). It is also much more efficient in terms of memory usage as reclamation is more targeted. Reclamation is naturally parallelized and balances across threads due to the nature of reference counting. Using `seize` *significantly* improves flurry's performance. Here's a read-heavy benchmark that was run on a 8 core machine, pinning before every operation:

## Remaining Items

- [ ] Determine if the `&'g Guard<'g>` stored by iterators is problematic.
- [ ] Potentially loosen `K/V: 'static` bounds now that there is no global collector.
- [ ] Fix `check_guard`. This is difficult because `seize::Collector` contains no allocations that can be easily accessed to check pointer equality. Perhaps flurry should expose it's own guard type that also contains an `&HashMap`, or a dummy allocation?


jonhoo:
This is so cool, thanks for taking it on!

jonhoo:
https://github.com/jonhoo/flurry/pull/103 should take care of that :+1:

jonhoo:
There — a rebase should do it

ibraheemdev:
Looks like you need to approve after every push...

ibraheemdev:
Everything is passing except Miri (need to disable isolation because seize calls num_cpus::get) and MSRV (which I can fix) :tada:

Looks like you need to approve again. I think you can [disable the approval requirement here in the settings](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-required-approval-for-workflows-from-public-forks), although Github should recognize that you already approved my branch...

ibraheemdev:
Can you approve the workflow again :upside_down_face: I believe I fixed the miri issue.

src/map.rs
```diff
@@ -3093,11 +3093,14 @@ where
     S: BuildHasher + Clone,
 {
     fn clone(&self) -> HashMap<K, V, S> {
-        let cloned_map = Self::with_capacity_and_hasher(self.len(), self.build_hasher.clone());
+        let cloned_map = Self::with_capacity_and_hasher(self.len(), self.build_hasher.clone())
+            .with_collector(self.collector.clone());
```
jonhoo, line 3097: Does this create a new collector, or a new handle to the same collector? If it's the latter, then this would preclude us from allowing `K` and `V` to be `!: 'static`.

tests/basic.rs
```diff
@@ -393,10 +393,9 @@ fn different_size_maps_not_equal() {
     let map1 = HashMap::<usize, usize>::new();
     let map2 = HashMap::<usize, usize>::new();
     {
-        let guard = map1.guard();
-        map1.insert(1, 0, &guard);
-        map1.insert(2, 0, &guard);
-        map2.insert(1, 0, &guard);
+        map1.pin().insert(1, 0);
```
jonhoo, line 396: Why these changes? This changes what the test actually tests a bit, since it no longer re-uses a single guard.

Cargo.toml
```diff
@@ -19,15 +19,12 @@ azure-devops = { project = "jonhoo/jonhoo", pipeline = "flurry", build = "15" }
...
-[features]
-sanitize = ['crossbeam-epoch/sanitize']
-
 [dependencies]
-crossbeam-epoch = "0.8.2"
 parking_lot = "0.10"
 num_cpus = "1.12.0"
 rayon = {version = "1.3", optional = true}
 serde = {version = "1.0.105", optional = true}
+seize = {git = "https://github.com/ibraheemdev/seize"}
```
jonhoo, line 27: When do you think we might get an actual release of `seize` we can depend on? I believe we can't publish to crates.io with a git dependency.

src/iter/mod.rs
```diff
@@ -10,12 +11,11 @@ use std::sync::atomic::Ordering;
...
     pub(crate) node_iter: NodeIter<'g, K, V>,
-    pub(crate) guard: &'g Guard,
+    pub(crate) guard: &'g Guard<'g>,
 }
 
-impl<'g, K, V> Iterator for Iter<'g, K, V> {
-    type Item = (&'g K, &'g V);
-    fn next(&mut self) -> Option<Self::Item> {
+impl<'g, K, V> Iter<'g, K, V> {
+    pub(crate) fn next_internal(&mut self) -> Option<(&'g K, &'g Linked<V>)> {
```
jonhoo, line 18: As far as I can tell the only use of this is in `impl Iterator for Iter`, so why separate it out?

src/iter/mod.rs
```diff
@@ -101,10 +107,10 @@ mod tests {
     fn values() {
         let map = HashMap::<usize, usize>::new();
 
-        let mut guard = epoch::pin();
+        let guard = map.guard();
         map.insert(1, 42, &guard);
         map.insert(2, 84, &guard);
-        guard.repin();
```
jonhoo, line 107: Ah, interesting, there's no notion of re-pinning to allow previously-protected garbage to be collected in `seize`?

src/iter/traverser.rs
```diff
@@ -26,11 +26,11 @@ pub(crate) struct NodeIter<'g, K, V> {
     /// Initial table size
     base_size: usize,
 
-    guard: &'g Guard,
+    guard: &'g Guard<'g>,
 }
 
 impl<'g, K, V> NodeIter<'g, K, V> {
-    pub(crate) fn new(table: Shared<'g, Table<K, V>>, guard: &'g Guard) -> Self {
+    pub(crate) fn new(table: Shared<'g, Table<K, V>>, guard: &'g Guard<'_>) -> Self {
```
jonhoo, line 33: Should this not be `'g` instead of `'_`?

src/map_ref.rs
```diff
@@ -28,7 +28,7 @@ impl<K, V, S> HashMap<K, V, S> {
     }
 
     /// Get a reference to this map with the given guard.
-    pub fn with_guard<'g>(&'g self, guard: &'g Guard) -> HashMapRef<'g, K, V, S> {
+    pub fn with_guard<'g>(&'g self, guard: &'g Guard<'_>) -> HashMapRef<'g, K, V, S> {
```
jonhoo, line 31:
```suggestion
    pub fn with_guard<'g>(&'g self, guard: &'g Guard<'g>) -> HashMapRef<'g, K, V, S> {
```
?

src/raw/mod.rs
```diff
@@ -75,16 +75,17 @@ impl<K, V> Table<K, V> {
     pub(crate) fn get_moved<'g>(
         &'g self,
         for_table: Shared<'g, Table<K, V>>,
-        guard: &'g Guard,
+        guard: &'g Guard<'_>,
```
jonhoo, line 78:
```suggestion
        guard: &'g Guard<'g>,
```
?

src/raw/mod.rs
```diff
@@ -75,16 +75,17 @@ impl<K, V> Table<K, V> {
...
         match self.next_table(guard) {
             t if t.is_null() => {
                 // if a no next table is yet associated with this table,
                 // create one and store it in `self.next_table`
-                match self.next_table.compare_and_set(
+                match self.next_table.compare_exchange(
                     Shared::null(),
                     for_table,
                     Ordering::SeqCst,
+                    Ordering::SeqCst,
```
jonhoo, line 88: I _think_ the fallback here can be `Relaxed`

src/raw/mod.rs
```diff
@@ -297,35 +298,29 @@ impl<K, V> Table<K, V> {
...
 
     #[inline]
-    pub(crate) fn bin<'g>(&'g self, i: usize, guard: &'g Guard) -> Shared<'g, BinEntry<K, V>> {
+    pub(crate) fn bin<'g>(&'g self, i: usize, guard: &'g Guard<'_>) -> Shared<'g, BinEntry<K, V>> {
         self.bins[i].load(Ordering::Acquire, guard)
     }
 
     #[inline]
     #[allow(clippy::type_complexity)]
-    pub(crate) fn cas_bin<'g, P>(
```
jonhoo, line 306: I think the `P` here was to make it easier to stick in both shared and owned, but I suppose that's not terribly important to retain.

src/reclaim.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+pub(crate) struct CompareExchangeError<'g, T> {
+    pub(crate) current: Shared<'g, T>,
+    pub(crate) new: Shared<'g, T>,
+}
+
+pub(crate) struct Shared<'g, T> {
+    ptr: *mut Linked<T>,
+    _g: PhantomData<&'g ()>,
```
jonhoo, line 88: If `seize::Linked` doesn't require the `'g`, should we just get rid of it from `Shared` as well?

src/reclaim.rs
```diff
@@ -0,0 +1,186 @@
...
+impl<T> From<*mut Linked<T>> for Shared<'_, T> {
+    fn from(ptr: *mut Linked<T>) -> Self {
+        Shared {
+            ptr,
+            _g: PhantomData,
+        }
+    }
+}
+
+impl<T> From<*const Linked<T>> for Shared<'_, T> {
```
jonhoo, line 146: I think I'd prefer to _not_ have this conversion, and have callers have to explicitly do the case from `*const`, since it _might_ lead to `&` -> `*const` -> `*mut` -> `Box`, which would be bad.

src/reclaim.rs
```diff
@@ -0,0 +1,186 @@
...
+    unsafe fn retire_shared<T>(&self, shared: Shared<'_, T>);
+}
+
+impl RetireShared for Guard<'_> {
+    unsafe fn retire_shared<T>(&self, shared: Shared<'_, T>) {
+        self.retire(shared.ptr, seize::reclaim::boxed::<T>);
+    }
+}
+
+pub(crate) unsafe fn unprotected() -> &'static Guard<'static> {
```
jonhoo, line 162: I'm a little surprised we can't use `Guard::unprotected` directly and need this helper?

src/node.rs
```diff
@@ -240,9 +241,8 @@ where
     /// Constructs a new bin from the given nodes.
     ///
     /// Nodes are arranged into an ordered red-black tree.
-    pub(crate) fn new(bin: Owned<BinEntry<K, V>>, guard: &Guard) -> Self {
+    pub(crate) fn new(bin: Shared<'_, BinEntry<K, V>>, guard: &Guard<'_>) -> Self {
```
jonhoo, line 244: The intention here was that you can only construct a `Node` from a `bin` you _know_ is initially owned. This new interface _allows_ someone to call `Node::new` with a thing that's _already_ shared, which I'm not sure is safe.

src/reclaim.rs
```diff
@@ -0,0 +1,186 @@
...
+    _g: PhantomData<&'g ()>,
+}
+
+impl<'g, T> Shared<'g, T> {
+    pub(crate) fn null() -> Self {
+        Shared::from(ptr::null_mut())
+    }
+
+    pub(crate) fn boxed(value: T, collector: &Collector) -> Self {
+        Shared::from(collector.link_boxed(value))
```
jonhoo, line 97: The fact that a ref to the `Collector` is needed to construct a `Shared` has me a little concerned — does _every_ seize-protected value hold a pointer to the collector? If so, that'll pretty significantly increase the size of a map where the key and value types are small. Say they're `u32`, but now we're sticking in a full `usize` pointer twice in every entry as well!

src/node.rs
jonhoo: Maybe it could just take a `BinEntry` directly and do the `Shared::new` bit internally?

src/node.rs
```diff
@@ -929,10 +935,12 @@ impl<K, V> TreeBin<K, V> {
     /// method.
     pub(crate) unsafe fn defer_drop_without_values<'g>(
         bin: Shared<'g, BinEntry<K, V>>,
-        guard: &'g Guard,
+        guard: &'g Guard<'_>,
     ) {
-        guard.defer_unchecked(move || {
-            if let BinEntry::Tree(mut tree_bin) = *bin.into_owned().into_box() {
+        guard.retire(bin.as_ptr(), |mut link| {
+            let bin = link.cast::<BinEntry<K, V>>();
```
jonhoo, line 941: I'm very surprised that `.cast` requires no `unwrap` or `unsafe` :thinking:

src/node.rs
```diff
@@ -929,10 +935,12 @@ impl<K, V> TreeBin<K, V> {
...
+
+            if let BinEntry::Tree(mut tree_bin) = Linked::into_inner(*Box::from_raw(bin)) {
```
jonhoo, line 943: Am I blind? Where's the `unsafe {}` for `Box::from_raw`?

jonhoo: Also, can't this use `bin.into_box()`?

jonhoo (changes requested):
This looks great! I had a few comments, but I don't _think_ any of them should be very hard to address :+1:

src/map.rs
ibraheemdev: It creates a new collector with the same settings.

tests/basic.rs
ibraheemdev: `map1` and `map2` have to use different guards, because they aren't tied to the same collector (which used to be the global one). It could do `let guard1 = map1.guard(); let guard2 = map2.guard();` though.

Cargo.toml
ibraheemdev: I can publish a release this weekend.

src/iter/mod.rs
ibraheemdev: It's used in `retain`, we need the `&Linked<V>` instead of `&V` in order to `retire`.

src/iter/mod.rs
ibraheemdev: No there isn't, because the reference counting is quite granular already.

src/iter/traverser.rs
ibraheemdev: I don't think so? We only care about the lifetime of the guard object, not the lifetime of the reference to the collector it holds.

src/map_ref.rs
ibraheemdev: Same as above I believe.

src/reclaim.rs
ibraheemdev: This ties the lifetime of the reference to that of the guard, which makes things a little safer. `Linked` is not equivalent to `Shared`, it's just a wrapper type with some extra metadata.

src/reclaim.rs
ibraheemdev: Yeah, this is a big footgun from crossbeam-epoch, I'll remove it.

src/reclaim.rs
ibraheemdev: I can remove it.

src/node.rs
ibraheemdev: Sure.

src/map.rs
jonhoo: Ah, okay, great!

src/iter/mod.rs
jonhoo: Ah, I missed that. All good then!

src/reclaim.rs
ibraheemdev: The map holds pointer to the `Linked<T>`, not the value directly, so it shouldn't increase the size?

src/node.rs
ibraheemdev: It is `unsafe` :thinking:

src/iter/mod.rs
jonhoo: Hmm, but if all borrows are tied to a given `&Guard`, then without repin a reader has no way to signal that reclaiming the protected objects are okay except to drop the `Guard` and grab a new one. Which, efficiency aside, can be fairly annoying and require something like `std::mem::replace`.

tests/basic.rs
jonhoo: Yeah, I was more thinking about the repeated `map.pin()`. Having two guards that each get reused for their respective maps seems fine.

src/reclaim.rs
jonhoo: It still increases the overall heap memory use of the map significantly. Previously we'd store (for each item):
```
sizeof(usize) + sizeof(K) + sizeof(usize) + sizeof(V)
```
whereas if each `Linked` _also_ holds a pointer to the container, and every `K` and `V` is a `Linked<_>`, then we're now storing
```
2 * sizeof(usize) + sizeof(K) + 2 * sizeof(usize) + sizeof(V)
```
which is a fairly significant increase in heap memory for a map when `K` and `V` are small, and there are many elements, such as maps to or from integers (which are _very_ common in practice).

src/node.rs
ibraheemdev:
```rust
unsafe {
    let _ = || {
        std::mem::uninitialized::<()>();
    };
}
```
compiles.

src/node.rs
jonhoo: But.. There's no `unsafe` block here? So how then can you call it?

src/node.rs
ibraheemdev: I'll change it to `into_box` :+1:

jonhoo: That's _wild_ that that compiles. I hate that so much. At the very least we should enable https://github.com/rust-lang/rust/issues/71668, though I know that doesn't fully fix this instance.

src/node.rs
jonhoo: Resolved in discussion below.

src/iter/mod.rs
ibraheemdev: Hm, that is true. I'll see if I can add a `repin` operation.

src/reclaim.rs
ibraheemdev: `Linked` actually stores a lot more than that. `Linked<()>` is 32 bytes, and packs a ton of metadata in there (2 unions). It stores a link to the head of the batch, the recorded epoch value when it was allocated, a link to the next node in the reservation list, the reference count and minimum recorded epoch of the batch, a link to the next node in the batch, and the reservation function :sweat_smile: All of this is pretty fundamental to the reclamation algorithm. Some of the data is inlined from what would instead be a thread-local list in crossbeam-epoch/hazard pointers. Do you think this is a big problem?

jonhoo: Oooooof, that's pretty painful. For _every_ value? That's definitely worrying. Consider a map with, say, 100k entries, all of which are `usize`s. Its heap memory use would increase three-fold (from 100k*(2*2*8)=3.2MB to 100k*(2*(8+32+8))=9.6MB). Now, whether it's a deal-breaker, that's harder to say... In its current state, flurry isn't going to be used since its performance is so poor, and at least with this its _performance_ is pretty good, even if its memory footprint is kind of lousy. That said, it would be in Java too, so maybe that's just what we signed up for :p

I think let's still go ahead with this, but maybe add a note about performance and memory characteristics to the docs, and highlight the expected memory use of the map given `N`, `sizeof(K)`, and `sizeof(V)`.

ibraheemdev: Hmmm.. it might be possible to initially only store the epoch value, and then only when a node is _retired_ allocate the 32 bytes needed to keep track of it's reservation status. That does mean we incur an extra pointer indirection anytime reference counts need to be incremented/decremented.

Do you think that would be worthwhile?

src/node.rs
ibraheemdev: Hm, it seems the bin might be null?

src/node.rs
ibraheemdev: We can't actually use `into_box`, because this is a `*mut Linked<T>`, not a `Shared<'_, T>`, but I did add the safety comments.

ibraheemdev: Enabling \`unsafe_op_in_unsafe_fn\` causes 34 new errors, do you think we can do it in a separate PR :sweat_smile:

tests/basic.rs
ibraheemdev: Fixed in d50737c467f47b6732057d36d29fbaa9a0751765.

src/raw/mod.rs
ibraheemdev: Fixed in 011a9df178983d95a9f6d43cfc404576904107be.

src/reclaim.rs
ibraheemdev: Fixed in 336477cb4cbd334fa216284459b21ed8b238c71a.

src/node.rs
jonhoo: Hehe, yep, that's fine :+1:

src/reclaim.rs
jonhoo: Does that mean an extra pointer indirection on every read? If so, I'd be curious to see the performance impact. But could very well be worth it, yeah!

src/node.rs
jonhoo: I'm not sure I follow?

src/reclaim.rs
ibraheemdev: No not on every read, just in retire and Guard::drop.

src/node.rs
ibraheemdev: Misunderstood how table bins were created, fixed in 386eaa75f33a0aab33646242525b45af5adf26ac.

src/reclaim.rs
jonhoo: That definitely sounds promising then!

src/iter/mod.rs
ibraheemdev: Added `flush` in ad9d5f29acce9a81b11f0e1875c39f45b6dae53e.

Cargo.toml
ibraheemdev: Released :tada:

ibraheemdev:
>  Determine if the &'g Guard<'g> stored by iterators is problematic.

I think this is fine, because we check that all guards come from the collector that we _own_, meaning that the map is guaranteed to outlive the inner 'g. Thus the lifetime of `Iter` is tied to the lifetime of the guard and map.

ibraheemdev:
> Potentially loosen K/V: 'static bounds now that there is no global collector.

All 'static bounds have been removed. Now that the collector is owned by the map, all retired values will be reclaimed when the map is dropped.

ibraheemdev:
I think the only blocker now is documentation. I'll open an issue about memory usage optimizations.

ibraheemdev:
Alright, I updated the public documentation, but there are still 100+ safety comments that talk about "thread X must be pinned to a lesser epoch" :sweat_smile: Do you think it's important all these comments are updated? They are all correct of course, the terminology is just wrong (s/pinned to an epoch/marked as active).

src/map.rs
```diff
@@ -3281,13 +3281,13 @@ mod tests {
 /// drop(r);
 /// ```
 ///
-/// # Keys and values must be static
+/// # Keys and values do not have be static
 ///
-/// ```compile_fail
+/// ```no_run
```
jonhoo, line 3286: Why make these `no_run`? Can't they just be regular doctests?

jonhoo:
> Do you think it's important all these comments are updated? They are all correct of course, the terminology is just wrong (s/pinned to an epoch/marked as active).

Unfortunately, yes, I do think it's important to update them, even if the terminology is just slightly off. Contributors _will_ read the safety comments and try to reason from them as to whether their modifications are correct, and if the comment doesn't match the semantics of the protection, it will be very confusing. Hopefully it'll be mostly search/replace though!

src/map.rs
ibraheemdev: True, I was just following the "get() can be non 'static test below".

ibraheemdev: Updated in 5db3eb82c3707fab6c788cd2d802f1e7288c2ce3.

ibraheemdev:
[Updated all the safety comments](c6078d6851091a217a0138137fae0ec752f8e68b), I think this can be merged now, provided I didn't make any typos :)

src/map.rs
```diff
@@ -1307,25 +1294,25 @@ where
         // there are three cases when a bin pointer is invalidated:
         //
         //  1. if the table was resized, bin is a move entry, and the resize has completed. in
-        //     that case, the table (and all its heads) will be dropped in the next epoch
-        //     following that.
+        //     that case, the table (and all its heads) will be retired.
```
jonhoo, line 1297: These changes aren't _quite_ right. This paragraph is reasoning about when in the future the bin pointer will actually be _reclaimed_. I _think_ it should read something like:

1. if the table was resized, bin is a move entry, and the resize has completed. in that case, the table (and all its heads) has already been retired.

src/map.rs
```diff
@@ -1307,25 +1294,25 @@ where
...
         //  2. if the table is being resized, bin may be swapped with a move entry. the old bin
-        //     will then be dropped in the following epoch after that happens.
+        //     will be retired.
```
jonhoo, line 1299: Same here:

2. if the table is being resized, bin may be swapped with a move entry. the old bin will only be retired after that happens.

src/map.rs
```diff
@@ -1307,25 +1294,25 @@ where
...
         //  3. when elements are inserted into or removed from the map, bin may be changed into
         //     or from a TreeBin from or into a regular, linear bin. the old bin will also be
-        //     dropped in the following epoch if that happens.
+        //     retired.
```
jonhoo, line 1302: retired only once that happens.

src/map.rs
```diff
@@ -1307,25 +1294,25 @@ where
...
+        // and the drop must happen _after_ we decrement the count (i.e drop our guard).
         let node = table.find(unsafe { bin.deref() }, h, key, guard);
         if node.is_null() {
             return None;
         }
-        // safety: we read the bin while pinning the epoch. a bin will never be dropped until the
-        // next epoch after it is removed. since it wasn't removed, and the epoch was pinned, that
-        // cannot be until after we drop our guard.
+
+        // safety: we loaded the bin while holding a guard, so any retirements
```
jonhoo, line 1313: _and_ any _future_ retirements must see us as active

src/map.rs
```diff
@@ -1524,7 +1511,7 @@ where
 
                             // free the node's value
                             // safety: any thread that sees this p's value must have read the bin before we stored null
-                            // into it above. it must also have pinned the epoch before that time. therefore, the
+                            // into it above. it must also have been marked as active. therefore, the
```
jonhoo, line 1514: _already_ been marked

src/map.rs
```diff
@@ -1711,13 +1697,11 @@ where
...
-            //     only queues a drop for the next epoch after removing the table.
+            //     guard, so it won't be dropped until after we drop that guard b/c this thread must
+            //     have been included in the reference count of a retirement.
             //
-            //  2. if table is read by init_table, then either we did a load, and the argument is
-            //     as for point 1. or, we allocated a table, in which case the earliest it can be
-            //     deallocated is in the next epoch. we are holding up the epoch by holding the
-            //     guard, so this deref is safe.
+            //  2. if table is read by init_table, we did so while holding a guard, so the
+            //     argument is as for point 1.
```
jonhoo, line 1704: This is missing the "or" part of the argument now.

src/map.rs
```diff
@@ -1758,25 +1741,22 @@ where
                 }
             }
 
-            // slow path -- bin is non-empty
```
jonhoo, line 1759: why remove this?

src/map.rs
```diff
@@ -1758,25 +1741,22 @@ where
...
-
             // safety: bin is a valid pointer.
             //
             // there are three cases when a bin pointer is invalidated:
             //
             //  1. if the table was resized, bin is a move entry, and the resize has completed. in
-            //     that case, the table (and all its heads) will be dropped in the next epoch
-            //     following that.
+            //     that case, the table (and all its heads) will be retired.
```
jonhoo, line 1749: Same notes as for this list further up.

src/map.rs
```diff
@@ -1817,18 +1796,17 @@ where
                     let mut p = bin;
 
                     old_val = loop {
-                        // safety: we read the bin while pinning the epoch. a bin will never be
-                        // dropped until the next epoch after it is removed. since it wasn't
-                        // removed, and the epoch was pinned, that cannot be until after we drop
-                        // our guard.
+                        // safety: we loaded the bin while holding a guard, so any retirements
```
jonhoo, line 1799: same note re _future_ retirements

src/map.rs
```diff
@@ -1945,15 +1923,15 @@ where
...
+                            //    they must have read it before the call to swap while
+                            //    marked as active (holding a guard), and are included in
+                            //    the reference count. therefore t won't be freed until _after_
+                            //    it decrements the reference count, which can only happen
+                            //    when that thread drops its guard, and with it, any reference
+                            //    to the value.
                             //  - another thread is about to get a reference to this value.
                             //    they execute _after_ the swap, and therefore do _not_ get a
-                            //    reference to now_garbage (they get `value` instead). there are
+                            //    reference to now_garbage (they get value instead). there are
```
jonhoo, line 1934: why remove the backtics?

src/map.rs
```diff
@@ -2087,18 +2061,17 @@ where
             // there are three cases when a bin pointer is invalidated:
             //
             //  1. if the table was resized, bin is a move entry, and the resize has completed. in
-            //     that case, the table (and all its heads) will be dropped in the next epoch
-            //     following that.
+            //     that case, the table (and all its heads) will be retired.
```
jonhoo, line 2064: same notes as for the earlier instance of this argument

src/map.rs
```diff
@@ -2194,18 +2166,19 @@ where
...
-                                //    they must have read it before the call to store_bin.
-                                //    because of this, that thread must be pinned to an epoch <=
-                                //    the epoch of our guard. since the garbage is placed in our
-                                //    epoch, it won't be freed until the _next_ epoch, at which
-                                //    point, that thread must have dropped its guard, and with it,
-                                //    any reference to the value.
+                                //  - another thread already has a reference to now_garbage.
+                                //    they must have read it before the call to swap while
+                                //    marked as active (holding a guard), and are included in
+                                //    the reference count. therefore t won't be freed until _after_
```
jonhoo, line 2172: we really do have a _lot_ of instances of this same argument :p

src/map.rs
```diff
@@ -2194,18 +2166,19 @@ where
...
+                                //    it decrements the reference count, which can only happen
+                                //    when that thread drops its guard, and with it, any reference
+                                //    to the value.
                                 //  - another thread is about to get a reference to this value.
-                                //    they execute _after_ the store_bin, and therefore do _not_ get a
-                                //    reference to the old value. there are no other ways to get to a
-                                //    value except through its Node's `value` field (which is now gone
-                                //    together with the node), so freeing the old value is fine.
+                                //    they execute _after_ the swap, and therefore do _not_ get a
+                                //    reference to now_garbage (they get value instead). there are
```
jonhoo, line 2178: why remove the backtics?

src/map.rs
```diff
@@ -2484,18 +2456,17 @@ where
             // there are three cases when a bin pointer is invalidated:
             //
             //  1. if the table was resized, bin is a move entry, and the resize has completed. in
-            //     that case, the table (and all its heads) will be dropped in the next epoch
-            //     following that.
+            //     that case, the table (and all its heads) will be retired.
```
jonhoo, line 2459: see comments on earlier instance of this argument

src/map.rs
```diff
@@ -2511,11 +2486,11 @@ where
...
                         // bin is valid at the time we look up bin in the table, at which point
-                        // the epoch is pinned by our guard. Since we found the next pointer
-                        // in a valid map and it is not null (as checked above and below), the
-                        // node it points to was present (i.e. not removed) from the map in the
-                        // current epoch. Thus, because the epoch cannot advance until we release
-                        // our guard, e is also valid if it was obtained from a next pointer.
+                        // we held a guard. Since we found the next pointer in a valid map and
+                        // it is not null (as checked above and below), the node it points to was
+                        // present (i.e. not removed) from the map while we were marked as active.
+                        // Thus, e cannot be dropped until we release our guard, e is also valid
```
jonhoo, line 2492:
```suggestion
                        // Thus, e cannot be dropped until we release our guard, and so e is also valid
```

src/map.rs
```diff
@@ -2783,14 +2753,14 @@ where
                     let mut head = Shared::null();
                     let mut tail = Shared::null();
                     while !e.is_null() {
-                        // safety: either `e` is `bin`, in which case it is valid due to the above,
```
jonhoo, line 2778: why remove the backtics? I'd rather add more of them in than take them out if this is to make the comment more consistent.

src/map.rs
```diff
@@ -2775,14 +2753,14 @@ where
...
-                        // present (i.e. not removed) from the map in the current epoch. Thus,
-                        // because the epoch cannot advance until we release our guard, `e` is also
-                        // valid if it was obtained from a next pointer.
+                        // safety: either e is bin, in which case it is valid due to the above,
+                        // or e was obtained from a next pointer. Any next pointer obtained from
+                        // bin is valid at the time we look up bin in the table, at which point
+                        // we held a guard. Since we found the next pointer in a valid map and
+                        // it is not null (as checked above and below), the node it points to was
+                        // present (i.e. not removed) from the map while we were marked as active.
+                        // Thus, e cannot be dropped until we release our guard, e is also valid
```
jonhoo, line 2762:
```suggestion
                        // Thus, e cannot be dropped until we release our guard, and so e is also valid
```

src/node.rs
```diff
@@ -422,18 +422,17 @@ impl<K, V> TreeBin<K, V> {
         // there are three cases when a bin pointer is invalidated:
         //
         //  1. if the table was resized, bin is a move entry, and the resize has completed. in
-        //     that case, the table (and all its heads) will be dropped in the next epoch
-        //     following that.
+        //     that case, the table (and all its heads) will be retired.
```
jonhoo, line 425: same note as for the other instances of this list

jonhoo (changes requested):
Mostly minor nits for the updated comments. I think we're almost good to go!

src/map.rs
ibraheemdev: The rest of the comment says "e is also valid if it was obtained from a next pointer." Maybe change it to a period instead of a comma?

src/map.rs
ibraheemdev: Fixed in f24162f1abd3f954060eff5026e961a95a876231

src/map.rs
ibraheemdev: Fixed in f24162f1abd3f954060eff5026e961a95a876231

src/map.rs
ibraheemdev: Fixed in f24162f1abd3f954060eff5026e961a95a876231

src/map.rs
ibraheemdev: Fixed.

src/map.rs
ibraheemdev: Fixed.

src/node.rs
ibraheemdev: Fixed.

src/map.rs
```diff
@@ -760,16 +738,15 @@ where
...
                     // next, let's talk about threads with _existing_ references to now_garbage.
                     // such a thread must have gotten that reference before the call to swap.
-                    // because of this, that thread must be pinned to an epoch <= the epoch of our
-                    // guard (since our guard is pinning the epoch). since the garbage is placed in
-                    // our epoch, it won't be freed until the _next_ epoch, at which point, that
-                    // thread must have dropped its guard, and with it, any reference to the value.
-                    unsafe { guard.defer_destroy(now_garbage) };
+                    // because of this, that thread must have been marked as active, and included
+                    // in the reference count, meaning the garbage will not be freed until
+                    // the thread drops its guard at the earliest.
```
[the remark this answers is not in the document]
ibraheemdev: Fixed.

src/map.rs
```diff
@@ -2060,13 +2036,11 @@ where
...
-            //     only queues a drop for the next epoch after removing the table.
+            //     guard, so it won't be dropped until after we drop that guard b/c this thread must
+            //     have been included in the reference count of a retirement.
             //
-            //  2. if table is read by init_table, then either we did a load, and the argument is
-            //     as for point 1. or, we allocated a table, in which case the earliest it can be
-            //     deallocated is in the next epoch. we are holding up the epoch by holding the
-            //     guard, so this deref is safe.
+            //  2. if table is read by init_table, we did so while holding a guard, so the
+            //     argument is as for point 1.
```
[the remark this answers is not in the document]
ibraheemdev: Fixed in cc93a0c01bf622dae4280130c4999b135ba4d05e.

src/map.rs
ibraheemdev: Fixed in cc93a0c01bf622dae4280130c4999b135ba4d05e.

src/map.rs
ibraheemdev: Yeah :smile:

ibraheemdev:
I think I've addressed all your comments.

src/map.rs
```diff
@@ -2492,7 +2492,7 @@ where
                         // we held a guard. Since we found the next pointer in a valid map and
                         // it is not null (as checked above and below), the node it points to was
                         // present (i.e. not removed) from the map while we were marked as active.
-                        // Thus, e cannot be dropped until we release our guard, e is also valid
+                        // Thus, e cannot be dropped until we release our guard. e is also valid
```
jonhoo, line 2497: Hmm, actually, the two instances of this now don't actually say _why_ `e` is valid if it was obtained from a next pointer. I _think_ the argument is the same as what comes before, which is why this used to be a `,`?

ibraheemdev: Ah I understand it now. You were right initially, `"so e is also valid ..." makes more sense.

ibraheemdev: fe8a171148ad6faacfe8d6b791ae93d9ca263905.

jonhoo:
Also, I would love to have you as a co-maintainer of this crate if you're up for it! You certainly know the code-base well enough at this point :sweat_smile:

jonhoo:
Just sent you a collaborator invite!

PR #106, opened 2022-03-25
Benchmark fix

jimvdl:
Benchmarks are still using `epoch` which was recently removed since `seize` was adopted for garbage collection. I needed to fix this in order for me to make some benchmarks for the concurrent counter.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/106)


jonhoo (approved):
Thanks!

PR #110, opened 2022-10-04
Fix references to Java source files

NotStirred:
Updates outdated references to source files.

Previously the link in `src/lib.rs` didn't include the revision, while `README.md` did. Was this intentional? 7fc19d97538434499ce93aba9e695eada69fd89c updates only the revision

It may be better to update the links to [an archive.org snapshot](https://web.archive.org/web/20221004094436/https://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/jsr166/src/main/java/util/concurrent/ConcurrentHashMap.java?revision=1.323&view=markup) instead

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/110)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
 dual licensed as above, without any additional terms or conditions.
 
   [source file from JSR166]: http://gee.cs.oswego.edu/dl/concurrency-interest/index.html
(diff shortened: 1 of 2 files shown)


jonhoo (approved):
Nope, that was not intentional, thanks for catching it!

I'd like to continue pointing to the canonical source — even if we have to update it occasionally, I think that's a better experience for anyone who does follow the links.

PR #114, opened 2023-03-12
Move to common CI

jonhoo:
https://github.com/jonhoo/rust-ci-conf

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/114)


jonhoo:
There miri check was already failing (we should probably fix that).

PR #123, opened 2024-03-12
Update `seize` to 0.3

ibraheemdev:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/flurry/123)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1664,7 +1664,7 @@ where
                 not_inserted,
             } => Err(TryInsertError {
                 current,
-                not_inserted: Linked::into_inner(*not_inserted),
+                not_inserted: not_inserted.value,
             }),
             PutResult::Inserted { new } => Ok(new),
             PutResult::Replaced { .. } => {
@@ -1731,9 +1731,7 @@ where
                     Err(changed) => {
                         assert!(!changed.current.is_null());
                         bin = changed.current;
-                        if let BinEntry::Node(node) =
-                            Linked::into_inner(*unsafe { changed.new.into_box() })
-                        {
+                        if let BinEntry::Node(node) = unsafe { changed.new.into_box() }.value {
                             key = node.key;
                         } else {
                             unreachable!("we declared node and it is a BinEntry::Node");
@@ -3317,7 +3315,7 @@ fn no_replacement_return_val() {
             map.put(42, String::from("world"), true, &guard),
             PutResult::Exists {
                 current: &String::from("hello"),
-                not_inserted: Box::new(map.collector.link(String::from("world"))),
+                not_inserted: Box::new(map.collector.link_value(String::from("world"))),
             }
         );
     }
--- a/src/map.rs
+++ b/src/map.rs
@@ -1020,15 +1020,11 @@ where
                         // the new bin will also be a tree bin. if both the high
                         // bin and the low bin are non-empty, we have to
                         // allocate a new TreeBin.
-                        Shared::boxed(
-                            BinEntry::Tree(TreeBin::new(
-                                // safety: we have just created `low` and its `next`
-                                // nodes and have never shared them
-                                unsafe { low.into_box() },
-                                guard,
-                            )),
-                            &self.collector,
-                        )
+                        // safety: we have just created `low` and its `next` nodes using `Shared::boxed`
+                        // and have never shared them
+                        let low_bin = unsafe { BinEntry::Tree(TreeBin::new(low, guard)) };
+
+                        Shared::boxed(low_bin, &self.collector)
                     } else {
                         // if not, we can re-use the old bin here, since it will
                         // be swapped for a Moved entry while we are still
@@ -1048,15 +1044,11 @@ where
                         unsafe { TreeBin::drop_tree_nodes(high, false, guard) };
                         high_linear
                     } else if low_count != 0 {
-                        Shared::boxed(
-                            BinEntry::Tree(TreeBin::new(
-                                // safety: we have just created `high` and its `next`
-                                // nodes and have never shared them
-                                unsafe { high.into_box() },
-                                guard,
-                            )),
-                            &self.collector,
-                        )
+                        // safety: we have just created `high` and its `next` nodes using `Shared::boxed`
+                        // and have never shared them
+                        let high_bin = unsafe { BinEntry::Tree(TreeBin::new(high, guard)) };
+
+                        Shared::boxed(high_bin, &self.collector)
                     } else {
                         reused_bin = true;
                         // since we also don't use the created low nodes here,
@@ -2793,18 +2785,11 @@ where
                         tail = new_tree_node;
                         e = e_deref.next.load(Ordering::SeqCst, guard);
                     }
-                    tab.store_bin(
-                        index,
-                        Shared::boxed(
-                            BinEntry::Tree(TreeBin::new(
-                                // safety: we have just created `head` and its `next`
-                                // nodes and have never shared them
-                                unsafe { head.into_box() },
-                                guard,
-                            )),
-                            &self.collector,
-                        ),
-                    );
+
+                    // safety: we have just created `head` and its `next` nodes using `Shared::boxed`
+                    // and have never shared them
+                    let head_bin = unsafe { BinEntry::Tree(TreeBin::new(head, guard)) };
+                    tab.store_bin(index, Shared::boxed(head_bin, &self.collector));
                     drop(lock);
                     // make sure the old bin entries get dropped
                     e = bin;
--- a/src/node.rs
+++ b/src/node.rs
@@ -241,9 +241,12 @@ where
     /// Constructs a new bin from the given nodes.
     ///
     /// Nodes are arranged into an ordered red-black tree.
-    pub(crate) fn new(bin: Box<Linked<BinEntry<K, V>>>, guard: &Guard<'_>) -> Self {
+    ///
+    /// # Safety
+    ///
+    /// The `bin` pointer was created with `Shared::boxed` and never shared.
+    pub(crate) unsafe fn new(bin: Shared<'_, BinEntry<K, V>>, guard: &Guard<'_>) -> Self {
         let mut root = Shared::null();
-        let bin = Shared::from(Box::into_raw(bin));
 
         // safety: We own the nodes for creating this new TreeBin, so they are
         // not shared with another thread and cannot get invalidated.
--- a/src/node.rs
+++ b/src/node.rs
@@ -2,7 +2,7 @@ use crate::raw::Table;
 use crate::reclaim::{Atomic, Collector, Guard, RetireShared, Shared};
 use core::sync::atomic::{AtomicBool, AtomicI64, Ordering};
 use parking_lot::Mutex;
-use seize::Linked;
+use seize::{Link, Linked};
 use std::borrow::Borrow;
 use std::thread::{current, park, Thread};
 
@@ -937,16 +937,16 @@ impl<K, V> TreeBin<K, V> {
         bin: Shared<'g, BinEntry<K, V>>,
         guard: &'g Guard<'_>,
     ) {
-        guard.retire(bin.as_ptr(), |mut link| {
+        guard.defer_retire(bin.as_ptr(), |link| {
             let bin = unsafe {
-                // SAFETY: `bin` is a `BinEntry<K, V>`
-                let ptr = link.cast::<BinEntry<K, V>>();
+                // SAFETY: `bin` is a `Linked<BinEntry<K, V>>`
+                let ptr: *mut Linked<BinEntry<K, V>> = Link::cast(link);
                 // SAFETY: `retire` guarantees that we
                 // have unique access to `bin` at this point
-                *Box::from_raw(ptr)
+                Box::from_raw(ptr).value
             };
 
-            if let BinEntry::Tree(mut tree_bin) = Linked::into_inner(bin) {
+            if let BinEntry::Tree(mut tree_bin) = bin {
                 tree_bin.drop_fields(false);
             } else {
                 unreachable!("bin is a tree bin");
@@ -985,7 +985,7 @@ impl<K, V> TreeBin<K, V> {
     ) {
         let mut p = from;
         while !p.is_null() {
-            if let BinEntry::TreeNode(tree_node) = Linked::into_inner(*p.into_box()) {
+            if let BinEntry::TreeNode(tree_node) = p.into_box().value {
                 // if specified, drop the value in this node
                 if drop_values {
                     let _ = tree_node.node.value.into_box();
@@ -1544,7 +1544,7 @@ mod tests {
         let entry = table.get_moved(table2, &guard);
         table.store_bin(0, entry);
         assert!(table
-            .find(&collector.link(BinEntry::Moved), 1, &2, &guard)
+            .find(&collector.link_value(BinEntry::Moved), 1, &2, &guard)
             .is_null());
         table.drop_bins();
         // safety: table2 is still valid and not accessed by different threads
@@ -1562,7 +1562,7 @@ mod tests {
         let entry = table.get_moved(table2, &guard);
         table.store_bin(0, entry);
         assert!(table
-            .find(&collector.link(BinEntry::Moved), 1, &2, &guard)
+            .find(&collector.link_value(BinEntry::Moved), 1, &2, &guard)
             .is_null());
         table.drop_bins();
         // safety: table2 is still valid and not accessed by different threads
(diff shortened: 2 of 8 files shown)


ibraheemdev:
It looks like the `Box` argument in `TreeBin::new` was making miri tag it as unique and making it mad when it was dereferenced again during the tree walk through a `prev` reference.

jonhoo:
Hmm, I _think_ what we're running into here is actually [this property](https://perso.crans.org/vanille/treebor/protectors.html#boxes-should-be-dereferencable-until-the-function-deallocates-them), where the expectation is that a `Box` argument to a function must be valid (and thus its backing memory not aliased) until it is deallocated. In our case, we don't deallocate it, we just convert it into a raw pointer with `Box::into_raw` and then we stick that raw pointer into the concurrent data structure so that readers can immediately start reading it. In other words, the memory behind the `Box` _is_ aliased during the execution of the function, which I think violates the LLVM `noalias` attribute.

Moving to passing in a raw pointer does avoid this problem since it means `noalias` isn't added, although it's pretty sad that it makes the whole function now be unsafe as the caller has to guarantee the pointer came from `Box::into_raw`.

What's _super_ interesting is that using the new tree borrows model `-Zmiri-tree-borrows` also makes Miri not complain (while keeping the `Box` in the argument type), and this makes me wonder whether I've either confused myself or whether it's just a missing bit in the current tree borrows implementation?

Would you mind splitting out the `seize` upgrade from this PR so that we can land that while continuing to dig on the Miri side of things?

---

I'm going to be bold and tag in @RalfJung and @Vanille-N here, just in case they can shed some light on the situation. Very briefly, the question is: is it okay to have a function that does this:
https://github.com/jonhoo/flurry/blob/48ff2e625498b1de9ec07c60e86e6bc18056d87b/src/node.rs#L244-246

and then immediately makes the  `*mut` that comes out of `Box::into_raw` available to other threads? [This Miri failure](https://github.com/jonhoo/flurry/actions/runs/7861392505/job/21449679449#step:5:166) suggests it isn't, and so does my read of [this paragraph in the tree borrows doc](https://perso.crans.org/vanille/treebor/protectors.html#boxes-should-be-dereferencable-until-the-function-deallocates-them) (since the `Box` isn't freed), but with `-Zmiri-tree-borrows` Miri no longer complains about the code in question.

Vanille-N:
> In other words, the memory behind the `Box` _is_ aliased during the execution of the function, which I think violates the LLVM `noalias` attribute.

Unless I misunderstand, the situation you describe isn't actually "aliased memory". At least it definitely doesn't violate `noalias` w.r.t. the original `Box` pointer because the pointer that is being shared is a child of that one and thus there is *not* a protected tag (the original `Box` pointer) for which a child (the `*mut` obtained) and a cousin (doesn't exist) are simultaneously accessed.
`noalias` does *not* mean "you cannot duplicate this pointer".

That doesn't answer why SB complains about the code, but at least (unless `Box::into_raw` does some weird stuff with the pointer ?) I don't think TB is expected to forbid this code.

jonhoo:
Thank you, that's helpful! You are indeed right, we never create cousins of the `*mut` (i.e., other children of the `Box`), only children of it. In a layman's sense I would certainly describe the memory as aliased the moment we share the `*mut`, but if LLVM's `noalias` only applies to the pointer while held by the `Box`, then we're okay. It just bends my brain that sharing the `*mut` _isn't_ a violation of `noalias` on the `Box` since it points to the same memory :sweat_smile: 

Yeah, no idea why SB and TB disagree here.

@ibraheemdev I think I'd recommend undoing the `Box` -> `*mut` change then, and instead adding `-Zmiri-tree-borrows` to `MIRIFLAGS`. It's listed as "very unstable", but I think it's better to keep the current code to get back to a clean Miri while accepting some potential for false negatives.

RalfJung:
> In our case, we don't deallocate it, we just convert it into a raw pointer with Box::into_raw and then we stick that raw pointer into the concurrent data structure so that readers can immediately start reading it. In other words, the memory behind the Box is aliased during the execution of the function, which I think violates the LLVM noalias attribute.

As long as the other threads are using the pointer that was created by `into_raw`, it's fine. `noalias` means there can't be any "outside" aliases, but if the function decides to share its noalias pointer with everyone, then everyone can access it. It's just crucial that they use a pointer that is "derived from" (in the provenance sense) the noalias pointer.

>  is it okay to have a function that does this:

Yes that seems totally fine.

The error indicates that you are using a pointer that is *not* derived from that `bin` argument, and that is the actual issue.

ibraheemdev:
Yeah, the issue that we are actually aliasing pointers. The outside is constructing a tree and giving it's children back pointers to the root node. It's then turning the root pointer into a Box and passing it to `new`. `new` traverses the tree and accesses to the back pointer to the root, which is aliased, as Miri flags.

@jonhoo I'm not sure we should be switching to tree borrows semantics because once the tree is constructed the pointers *are* aliased, so we shouldn't really be using `Box` again.

jonhoo:
Ahhh, yes, you're right @ibraheemdev. Although the problem isn't the aliasing (I think we're actually doing that part right), it's the provenance. And concurrency isn't part of the problem. To give a bit more detail for Ralf/Neven since this is interesting (and possibly just a shortcoming of SB):

The heap allocated value is originally constructed here:

https://github.com/jonhoo/flurry/blob/48ff2e625498b1de9ec07c60e86e6bc18056d87b/src/map.rs#L2780-L2781

This happens as part of [constructing a doubly-linked list just prior to the call with the `Box` argument](https://github.com/jonhoo/flurry/blob/48ff2e625498b1de9ec07c60e86e6bc18056d87b/src/map.rs#L2758-L2797). That code all deals just in `*mut`s. The doubly-linked list is not shared to other threads at this point.

On the first iteration through, `head` and `tail` are both set to point to that allocation, with their provenance set as children of the original allocation. Then, when the next element is added to the linked list, the `prev` pointer of that element is set to the current `tail` pointer, which is == `head`. From a provenance perspective, this is then a child of `tail` (which is a sibling(?) of `head`). This assignment happens here:

https://github.com/jonhoo/flurry/blob/48ff2e625498b1de9ec07c60e86e6bc18056d87b/src/map.rs#L2779

After the linked list construction, we now turn `head` back into a `Box`:

https://github.com/jonhoo/flurry/blob/48ff2e625498b1de9ec07c60e86e6bc18056d87b/src/map.rs#L2801-L2806

From an aliasing perspective, not a problem — we have another raw pointer to it, but as long as it isn't dereferenced, things are fine. The `Box` has provenance as a child of `head`.

We then pass that `Box` to `TreeBin::new`, which immediately calls `Box::into_raw` on it again (it only takes an argument of type `Box` to enforce that the caller has indeed heap-allocated the value). The resulting pointer has provenance as a child of the `Box`. The function then does "some stuff" to turn the linked list into a tree instead (still no concurrent access). The pointer that's a child of the `Box` gets assigned as the root. We then walk the linked list (forwards), assigning "left" and "right" pointers using the pointers that come out of next/prev pointers. Meaning the root now has provenance as the child of the `Box`, but all the _other_ pointers in the tree have provenance from the linked list construction.

Once we now go to _use_ this tree for anything, we hit the problem (at least for SB). I think what happens is that Miri sees the (read) access of the root, and concludes that this is access via the provenance of the `Box`. So, the provenance of that `Box` must be "alive" (I don't know the right term). But then it sees another (read) access through the `.prev` pointer of a node in the tree, that that pointer has provenance from the original `head` (well, `tail` really). And since that provenance is "before" the `Box` provenance, SB concludes that the access to `head` invalidates the access via `Box`.

Why TB allows this I'm not entirely sure, but it sounds like you're onto something @RalfJung in that it may have sufficient information about the provenance tree here to conclude that it's okay for these provenances to co-exist when all the accesses are reads?

---

In terms of where we go from here, I suppose I'm okay switching to the `*mut` argument type and just documenting clearly in both the method signature and all calls that the argument must be/is a value constructed via `Shared::boxed`.

RalfJung:
> Although the problem isn't the aliasing (I think we're actually doing that part right), it's the provenance.

There's no difference between these. Provenance is how aliasing is defined.

If you have a long-lived `Box`, then because `Box` is `noalias`, all other long-lived pointers that are used while the `Box` is "live" must be *derived from* that `Box`. From your description it sounds like this is indeed not the case.

If there are truly only `Box` and no references anywhere here, you may may want to add your voice in https://github.com/rust-lang/unsafe-code-guidelines/issues/326 arguing for the `noalias` on `Box` to be removed.

> Why TB allows this I'm not entirely sure, but it sounds like you're onto something @RalfJung in that it may have sufficient information about the provenance tree here to conclude that it's okay for these provenances to co-exist when all the accesses are reads?

My theory is that in the testcase, everything is read-only. TB allows aliasing of `noalias` pointers as long as there are no writes. If you have other testcases where a similar pattern occurs, but then after the `Box` is created
- The original pointer is used for mutation
- Then a pointer derived from the `Box` are used for mutation

then there should be UB with TB as well.

jonhoo:
> > Although the problem isn't the aliasing (I think we're actually doing that part right), it's the provenance.
> 
> There's no difference between these. Provenance is how aliasing is defined. So I don't know what you mean by this.

I think it's probably a mistake to use these terms interchangeably. While we may define aliasing in terms of provenance, that doesn't mean they are synonyms. Semi-formally, at least in my head, provenance is "knowledge of ow we got here", while aliasing is "a statement about the provenance graph". But in this case, I was using the less-formal meaning of aliasing, namely "are two pointers pointing to the same address" (which yes, I know, is an overly-simple way of looking at the world 😅). So what I meant was that the problem isn't that we have two `*mut` to the allocation behind a `Box` while the `Box` lives (which was my read of @ibraheemdev's comment), it's that _after_ the whole `Box` dance, we now have `*mut` of different (and incompatible as per below) provenances to the underlying memory in our data structure.
 
> If you have a long-lived `Box`, then because `Box` is `noalias`, all other long-lived pointers that are used while the `Box` is "live" must be _derived from_ that `Box`. From your description it sounds like this is indeed not the case.
> 
> If there are truly only `Box` and no references anywhere here, you may may want to add your voice in [rust-lang/unsafe-code-guidelines#326](https://github.com/rust-lang/unsafe-code-guidelines/issues/326) arguing for the `noalias` on `Box` to be removed.

Yes, I think that's right. The various "states of the world" are, in order:

1. A `Box` is created.
2. The `Box` is turned into a `*mut` with `Box::into_raw`.
3. The `*mut` from 2 is stored in two places.
4. A `Box` is re-created from the first of the `*mut` from 3. The second `*mut` is not dereferenced while this `Box` lives.
5. A function is called, and passed the `Box` from 4 as an argument.
6. The function immediately turns that `Box` into a `*mut` with `Box::into_raw`.
7. The `*mut` from 6 and the second `*mut` from 3 are both dereferenced (read-only).
8. (eventually) One of the `*mut` from 7 is turned back into a `Box` and dropped. This happens only once we guarantee that none of the other `*mut` will ever be dereferenced again.

I think the question is whether this means "the `Box` can't be `noalias`". I suppose in the execution above, `Box::from_raw` could technically (and legally) do something like relocate the allocation, which would break this entire flow. Which suggests that the mere existence of a `Box` implies any alias to the allocation cannot be relied upon any more. Which is indeed what `noalias` implies and what SB tells us. So I think I'm onboard with that (because I believe that's something that should be legal and safe for `Box` to do).

I suppose TB in this case allows it because it realizes that `Box::into_raw` and `Box::from_raw` _don't_ do that? Maybe?

> > Why TB allows this I'm not entirely sure, but it sounds like you're onto something @RalfJung in that it may have sufficient information about the provenance tree here to conclude that it's okay for these provenances to co-exist when all the accesses are reads?
> 
> My theory is that in the testcase, everything is read-only.

Yep, see above.

RalfJung:
> what I meant was that the problem isn't that we have two *mut to the allocation behind a Box while the Box lives (which was my read ***@***.***'s comment)

I would say that is exactly the problem. The point is that the Box is live as long as at least one pointer derived from it is live. That is a fundamental ingredient of both Stacked Borrows and Tree Borrows.

RalfJung:
> I suppose TB in this case allows it because it realizes that Box::into_raw and Box::from_raw don't do that? Maybe?

No, TB allows it because there are no writes.

jonhoo:
>> what I meant was that the problem isn't that we have two *mut to the allocation behind a Box while the Box lives
> I would say that is exactly the problem. The point is that the Box is live as long as at least one pointer derived from it is live. That is a fundamental ingredient of both Stacked Borrows and Tree Borrows.

Oh interesting, I hadn't thought about it that way. If that is indeed the rule, then yes, totally agree that we violate `noalias` since we _do_ have other live pointers to the allocation while a _child_ of the `Box` is also live 👍

>> I suppose TB in this case allows it because it realizes that Box::into_raw and Box::from_raw don't do that? Maybe?
> No, TB allows it because there are no writes.

Ah, yes, I suppose "relocating" is really just one example of a write, but any write indeed.

src/map.rs
```diff
@@ -1020,15 +1020,13 @@ where
...
-                                // nodes and have never shared them
-                                unsafe { low.into_box() },
-                                guard,
-                            )),
-                            &self.collector,
-                        )
+                        // safety: we have just created `low` and its `next`
+                        // nodes and have never shared them
+                        let low_bin = unsafe {
+                            BinEntry::Tree(TreeBin::new(Box::into_raw(low.into_box()), guard))
```
jonhoo, line 1026: Based on my read of the discussion, I think we shouldn't `Box::into_raw(low.into_box())` here (and elsewhere) — we shouldn't be re-constructing a `Box` at all. Instead, we should grab the raw pointer directly with `low.0.into_inner()` so that we preserve the right provenance. In fact, we should probably remove `fn into_box` entirely.

src/map.rs
```diff
@@ -1020,15 +1020,13 @@ where
...
-                            BinEntry::Tree(TreeBin::new(
-                                // safety: we have just created `low` and its `next`
-                                // nodes and have never shared them
-                                unsafe { low.into_box() },
-                                guard,
-                            )),
-                            &self.collector,
-                        )
+                        // safety: we have just created `low` and its `next`
+                        // nodes and have never shared them
```
jonhoo, line 1024: I'd like us to update this (and the other) safety comment to also reflect the new safety requirement of `TreeBin::new` (namely that the pointer in the first argument was created by a `Box`).

jonhoo (changes requested):

jonhoo:
> The point is that the Box is live as long as at least one pointer derived from it is live. That is a fundamental ingredient of both Stacked Borrows and Tree Borrows.

Follow-up question on this: how does this work when you have code like this:
```rust
let a = Box::new(42);
let b = Box::into_raw(a);
let c = unsafe { Box::from_raw(b) };
drop(c);
```
`c` is live, and is derived from `b`, which is in turn derived from `a`. Which by your statement implies that `a` is live. But that violates `noalias` since we then really have two "live" `Box`es to the same allocation?

RalfJung:
> Oh interesting, I hadn't thought about it that way. If that is indeed the rule, then yes, totally agree that we violate noalias since we do have other live pointers to the allocation while a child of the Box is also live 👍

Okay looks we found the cause of the misunderstanding then. :)
https://github.com/rust-lang/unsafe-code-guidelines/issues/450 is an issue tracking whether this is the right call. It's about `&mut` as that's where this happens most often, but also applies to `Box`.

> c is live, and is derived from b, which is in turn derived from a. Which by your statement implies that a is live. But that violates noalias since we then really have two "live" Boxes to the same allocation?

No, this is fine for the same reason reborrowing is fine with `&mut` (which in a sense also means there are aliases). `nolias`  is not violated precisely because of the parent-child relationship. It's only an aliasing violation when you have two live pointers that are not in a parent-child relationship. To be more precise, this is asymmetric: writing to a pointer means you can't use any of its `noalias` children or uncles any more, but you can still use all of its parents.

src/map.rs
ibraheemdev: I reworked `TreeBin::new` and the safety comments around `Shared::boxed` to avoid the `Shared->*mut T->Shared` conversions entirely.

ibraheemdev:
The Miri failures are unrelated and caused by a recent nightly regression (https://github.com/rust-lang/rust/pull/124013). Everything passes running locally an older Miri.

src/node.rs
```diff
@@ -241,9 +241,12 @@ where
     /// Constructs a new bin from the given nodes.
     ///
     /// Nodes are arranged into an ordered red-black tree.
-    pub(crate) fn new(bin: Box<Linked<BinEntry<K, V>>>, guard: &Guard<'_>) -> Self {
+    ///
+    /// # Safety
+    ///
+    /// The `bin` pointer was created with `Shared::boxed` and never shared.
```
jonhoo, line 247:
```suggestion
    /// The `bin` pointer and its successors were created with `Shared::boxed` and never shared.
```

jonhoo (approved):
Excellent, this looks good to me, thanks!

One interesting question here is whether we can now remove `into_box` entirely?

jonhoo:
Releasing in https://github.com/jonhoo/flurry/pull/124

